PR #14, opened 2020-05-27
Fix dropout issues

14thibea:
- fix dropout issues in parameters
- allow to change dropout in resnet model
- fix minor bug in SVM commandline


clinicadl/clinicadl/tools/deep_learning/models/slice_level.py
```diff
@@ -46,7 +46,7 @@ def resnet18(**kwargs):
         p.requires_grad = True
 
     # add a fc layer on top of the transfer_learning model and a softmax classifier
-    model.add_module('drop_out', nn.Dropout(p=0.8))
```
mdiazmel, line 49: It seems that dropout value was fixed by default to 0.8 (at least for the slice models). Can you confirm this? 
When adding this parameter to the cli, what is the default value? (I think that I see a 0 in this file `linicadl/clinicadl/tools/deep_learning/iotools.py `).

mdiazmel (commented):
Hi @14thibea ,
Thanks for updating your last PR. 
I have a small question/comment before merging (see below in the code review)...

clinicadl/clinicadl/tools/deep_learning/models/slice_level.py
14thibea: Hello @mdiazmel,

The default dropout value is different for the subject (0.5), patch (0) and slice-level (0.8).
I don't know if we should keep it like that (and how ?) or not.

mdiazmel: I think that the best is to keep the same values used in the publication (the option by default).

BTW, your previous PR modifies also the value of the dropout parameter for the other cases (patch and subject)? (I found that the dropout values are hard-coded in the model initialization, e.g., see `clinicadl/clinicadl/tools/deep_learning/models/subject_level.py`, line 17):
```python
class Conv5_FC3(nn.Module):
    """
    Classifier for a binary classification task

    Subject level architecture used on Minimal preprocessing
    """
    def __init__(self, dropout=0.5):
```

14thibea: Yes the previous PR changes the default value for subject and patch networks, I will propose a new commit to preserve the default dropout values for each case when the user specifies nothing.

mdiazmel:
I'll merge this PR. 
Thanks @14thibea !

PR #20, opened 2020-06-07
Group arguments of train in commandline

14thibea:
The goal of this PR is to group the arguments of train which can be very different depending of the type of the network.

REMOVED:
- `sampler` (unused in code)
- `pretrained_path` (replaced by `transfer_learning_path`)
- `hippocampus_roi` (replaced by `train` subparser)
- `network_type` (replaced by `patch` subparser)
- `use_gpu` (replaced by `use_cpu`)

ADDED:
- `roi` subparser of `train`
- `multicnn`subparser of `patch`

RENAMED:
- `pretrained_difference` --> `transfer_learning_difference`
- `mri_plane` --> `slice_direction`
- `subject` --> `image`
- `use_extracted_patches` --> `use_extracted_patches` / `use_extracted_slices`

CHANGED:
- The behaviour of `evaluation_steps` was slightly changed (default value=0 means there is no  evaluation during the epoch).
- `minmaxnormalization` was replaced by `unnormalize` (to have normalization by default).
- ROI extraction can only be on-the-fly (no prior extraction possible).
- Default behaviour is to try to run on GPU and to raise an Error if no GPU is available.

Also a suite of automatic test with pytest was added in `clinicadl/tests`.


14thibea:
Suggestions (that have not been implemented yet):
    - remove `num_cnn` for multi-patch and automatically find the right number of CNNs
    - remove `transfer_learning_multicnn` and find automatically if the transfer learning folder is multi or single.
- `visualization` could be implemented for 3D image autoencoder (only for patch now).
- By default, there is no normalization in image_based pipeline. It should be the contrary (default perform min-max normalization, else specify).

14thibea:
Other suggestions:
- By default use the outputs of `extract` (prepare_dl = True) ?
- By default use a GPU (and instead have a flag --cpu to force training on CPU)
- for patch and slice, if `split = None` all the splits of the k-fold CV are done. It could be implemented for subject level and default value for `split` could be set to None.
- `transfer_learning_path` is not very clear. For subject-level AE it is the path to the model file (filename included) and for others to the root folder of the experiment (the implementation of subject-level AE should be changed).
- Could subparsers of `patch` be `autoencoder`, `multi-cnn` and `single-cnn` instead of `autoencoder` and `cnn` ? This would make the used realize that these choices exist and that they are quite different (also some arguments are specific to multi-cnn)
- It could be also good to add a new subparser for `roi` instead of mixing them in `patch subparser` though they share the same core code.
- Preprocessings that can be chosen in the commandline are `linear` and `mni`. For more homogeneity `mni` could be renamed `extensive` or `t1-volume` (and `linear` would be `t1-linear`).

mdiazmel:
Thanks @14thibea for your PR!

I will transform your suggestions in issues, in this way we can have a kind of todo list for the next features to implement. We can continue the discussion concerning these suggestions in the issue's discussion.

Let me know when your PR is ready for review.

14thibea:
The PR is ready for review !

clinicadl/clinicadl/cli.py
```diff
@@ -145,51 +163,47 @@ def train_func(args):
...
+            split=args.split,
+            epochs=args.epochs,
+            batch_size=args.batch_size,
+            optimizer='Adam',
+            weight_decay=args.weight_decay,
+            dropout=args.dropout,
+            gpu=args.use_gpu,
+            num_workers=args.nproc,
+            selection_threshold=args.selection_threshold,
+            prepare_dl=args.prepare_dl
```
mdiazmel: I think that is necessary to change the name of this parameter (_cf._ issue #24 )

clinicadl/clinicadl/cli.py
```diff
@@ -3,11 +3,33 @@
 import argparse
 
 from clinicadl.tools.deep_learning.iotools import Parameters
+from colorama import Fore
```
mdiazmel, line 6: I'm not comfortable with the idea of having a new dependence for clinicadl (only to show nice colors). Well, at least we are coherent with Clinica...

clinicadl/clinicadl/subject_level/train_CNN.py
```diff
@@ -90,7 +88,9 @@ def train_cnn(params):
 
     # Define criterion and optimizer
     criterion = torch.nn.CrossEntropyLoss()
-    optimizer = eval("torch.optim." + params.optimizer)(filter(lambda x: x.requires_grad, model.parameters()), params.learning_rate, weight_decay=params.weight_decay)
+    optimizer = eval("torch.optim." + params.optimizer)(filter(lambda x: x.requires_grad, model.parameters()),
```
mdiazmel, line 92: Nice! It's always a nightmare to break these long lambda expressions!

mdiazmel (approved):
Hi @14thibea,

Thanks for your PR, nice job!

Some questions:
- With the new subparsers, the behavior of the json writer changes? I mean, if we have and old file `commandline_CNN.json`, it can be adapted easily to the renamed parameters?
- Did you tested the new subparsers with cli ? We have to do some modifications to the `test_cli.py` file, but it can be done after merging.

[edit] I forgot, there are some style issues to solve before merging. There are nice alternatives, depending on the editor that you use. I suggest you two tools that can be used from the command line:   [`pycodestyle`](https://github.com/PyCQA/pycodestyle) to find the errors and [`autopep8`](https://github.com/hhatto/autopep8) to fix them automatically.

clinicadl/clinicadl/cli.py
14thibea: I agree that we shouldn't add other dependencies, but as this dependency is already in clinica and that it allows us to have a similar command line aspect, I thought it was okay...

clinicadl/clinicadl/subject_level/train_CNN.py
14thibea: I will check that it was done everywhere

clinicadl/clinicadl/cli.py
mdiazmel: Yes, you're right, it's already in clinica. Let's keep it!

clinicadl/clinicadl/subject_level/train_CNN.py
14thibea: Done in https://github.com/aramis-lab/AD-DL/pull/20/commits/9d80f42b11ab2a0b87315d1c207cb59ed4319cdb

clinicadl/clinicadl/cli.py
14thibea: Done in https://github.com/aramis-lab/AD-DL/pull/20/commits/d5be1d34e8a59a65742f8870bd2a69f4e4ac6e6f

14thibea:
We could create a tool to transform a former `commandline.json` to a new one adapted to the latest version based on `read_json` (modified in https://github.com/aramis-lab/AD-DL/pull/20/commits/d768f90208bdb9ee0cf672982415de20c83a0eee).
Also the modification of `read_json` allow to use all the scripts of evaluation of image-based networks whatever the version, and it could be used in the future `classify` pipeline to ensure retro-compatibility.

14thibea:
I also added in the latest commits the ROI subparser and changed the names of preprocessings + default normalization for image-level networks.

clinicadl/clinicadl/cli.py
14thibea: Actually all the core codes (`slice_level/train_CNN.py`, `patch_level/train_singleCNN.py`, etc) use the variable `prepare_dl` (and not `use_extracted_patches`). I suggest that we rename this variable but only in commandline and that the names are converted in  `read_json` (see https://github.com/aramis-lab/AD-DL/pull/20/commits/4e184b05b27c6b40eced3236f5bda86de1d9bf67)

14thibea:
I didn't test the new commandline with the latest modifications, I will tell you when it is done !

clinicadl/clinicadl/main.py
```diff
@@ -2,27 +2,29 @@
...
     arguments = vars(args)
 
     if (arguments['task'] != 'preprocessing') \
             and (arguments['task'] != 'extract') \
             and (arguments['task'] != 'generate') \
             and (arguments['task'] != 'tsvtool'):
-        commandline_to_json(commandline, task_type)
+        commandline_to_json(commandline, arguments["mode_task"])
+
+        if args.gpu and not torch.cuda.is_available():
```
mdiazmel: The variable `args.gpu` doesn't exist anymore.
```suggestion
        if not args.use_cpu and not torch.cuda.is_available():
```

clinicadl/clinicadl/tools/deep_learning/data.py
```diff
@@ -303,7 +341,8 @@ def __init__(self, caps_directory, data_file, transformations=None, mri_plane=0,
         else:
             raise Exception('The argument datafile is not of correct type.')
 
-        # This dimension is for the output of image processing pipeline of Raw: 169 * 208 * 179
+        # This dimension is for the output of image processing pipeline of Raw:
+        # 169 * 208 * 179
         if mri_plane == 0:
             self.slices_per_patient = 169 - 40
```
mdiazmel, line 347: Warning: when the slices are extracted some of them are discarded. In a next PR we must modify this hard-coded value (40) by a variable.

mdiazmel (commented):
Hi @14thibea ,

I found a small bug on the main function when running the training on cpu. I suggest a change (see below).

I wonder why you changed the parameter `stride_size` to `patch_stride` and then you reverse this change? I agree with both but maybe `patch_stride` makes more sens... (it's only a detail). I let you choose but if the argument is that we must be coherent with Clinica's pipeline (that uses the same parameter), we can modify the pipeline.

clinicadl/clinicadl/tools/deep_learning/data.py
14thibea: I opened https://github.com/aramis-lab/AD-DL/issues/35 on this subject !

14thibea:
Hi @mdiazmel,

I didn't realize I changed the name of the variable several times... But I think `stride_size` is better as it is the name of the variable in clinica https://github.com/aramis-lab/clinica/wiki/DeepLearning_PrepareData.

mdiazmel:
Ok! 

> I didn't realize I changed the name of the variable several times... But I think `stride_size` is better as it is the name of the variable in clinica https://github.com/aramis-lab/clinica/wiki/DeepLearning_PrepareData.

That answer is enough for me.

I'll merge your PR. Thanks again!

alexandreroutier:
> Hi @mdiazmel,
> 
> I didn't realize I changed the name of the variable several times... But I think `stride_size` is better as it is the name of the variable in clinica https://github.com/aramis-lab/clinica/wiki/DeepLearning_PrepareData.

In that case, should we change the wording in `deeplearning-prepare-data` to stay coherent?

mdiazmel:
Hi @alexandreroutier ,

We kept `stride_size` at the end. Nothing to change.

Issue #15, opened 2020-05-27
`clinicadl classify` to implement


mdiazmel:
The `classify` task should be implemented.
Scripts examples for evaluation of the trained models can be found inside of each folder (e.g. `clinicadl/clinicadl/subject_level/evaluation.py`).

PR #34, opened 2020-06-11
Add classify task

mdiazmel:
This PR will add a simple way to perform inference on a single/multiple subject using the pretrained models.  It closes issue #15.

The command line should be like this:
```
clinicadl classify [-h] [-cpu] <caps_dir> <tsv_file> <model_path>
```
It writes the output files in the folder structure given by model_path unless an output_dir is specified (this apply only for the `image` mode. Other modes, `slice`, `patch`, `roi`, need the validation file with the models and output is always written in the default folder structure). There are options to use or not labels (by default it assumes that the dataset is labeled) and also to use extracted features or not.


clinicadl/clinicadl/classify/inference.py
```diff
@@ -1,26 +1,394 @@
...
+
+            # load the best trained model during the training
+            model, best_epoch = load_model(
+                model,
+                join(model_path, 'cnn-%i' % n),
+                gpu,
+                filename='model_best.pth.tar')
+
+            results_df, metrics = test(model, test_loader, gpu, criterion)
+            print("Patch level balanced accuracy is %f" % metrics['balanced_accuracy'])
```
14thibea: There might be a problem with the multi-CNN classification: one dataframe `results_df` is computed per CNN then they are not combined and only the last one is returned at the end of the function.

clinicadl/clinicadl/classify/inference.py
```diff
@@ -1,26 +1,394 @@
...
+
+def inference_from_patch_model(caps_dir, tsv_file, model_path, options):
+    '''
+    Inference using an image/subject CNN model
+
+
+
+    '''
+    from clinicadl.tools.deep_learning.data import MRIDataset_patch, MRIDataset_patch_hippocampus
+    from clinicadl.patch_level.utils import test
```
14thibea: This test function gives the label on the patch level. Maybe classify should aggregate to the image level ? This comment is also for slice | roi.

clinicadl/clinicadl/classify/inference.py
```diff
@@ -1,26 +1,394 @@
...
+    if hasattr(options, 'use_gpu'):
+        options.use_cpu = not options.use_gpu
+
+    if hasattr(options, 'use_extracted_patches'):
+        options.prepare_dl = not options.use_extracted_patches
+
+    if options.mode == "patch" and hasattr(options, "network_type"):
+        if options.network_type == "multi":
+            options.mode_task = "multicnn"
+        del options.network_type
```
14thibea: I think it would be easier to put all the constraints on retro-compatibility in `read_json` if it is possible.
Moreover I just realized that `read_json` is not supposed to read anything about the GPU option as it should be always specified and not dependent from the previous task.
This implies some changes in `read_json`.

14thibea (commented):
I had two main questions about the classify pipeline because I am not sure anymore what have been decided:
- can it be executed on tsv files which do not contain the true label ? (only a subject/session list)
- do we want to let the users the possibility to store the results where they want (and not in a folder inside the experiment folder) ? If they can freely choose they may lose the traceability between the file and how it was obtained, but it may be too much constraints and experiments folders may become heavy. I don't know what is better.

Maybe one way to avoid to lose the link between the model used and the tsv file inferred could be to generate a small commandline with the output file to remember how it was generated ?

mdiazmel:
Hi @14thibea ,
Thanks for your feedback.

> I had two main questions about the classify pipeline because I am not sure anymore what have been decided:

Nothing has been decided yet ;). Concerning your questions:

>  * can it be executed on tsv files which do not contain the true label ? (only a subject/session list)

My personal feeling is that the final goal of a diagnostic support system is to provide an output on unlabeled data. Of course, `clinicadl` is an experimental tool (and users should be aware of what that means), but we may leave the choice to the users. (If we decide to go this way, I need to change some parts of the code.)

>  * do we want to let the users the possibility to store the results where they want (and not in a folder inside the experiment folder) ? If they can freely choose they may lose the traceability between the file and how it was obtained, but it may be too much constraints and experiments folders may become heavy. I don't know what is better.

I'm agree that the choice is not easy. Concerning the traceability, users can argue in a simply manner that a X or Y pretrained-model was used to find the results (traceability of models is guarantee with the json files). Another approach can be to propose some kind of BIDS/CAPS output structure to store results... maybe @alexandreroutier can give us some ideas.

clinicadl/clinicadl/classify/inference.py
mdiazmel: Done in 03c24f2!

clinicadl/clinicadl/classify/inference.py
mdiazmel: It seems that this is solved using the new `test` function (after the refactoring of PR #42 )

clinicadl/clinicadl/classify/inference.py
mdiazmel: Solved using the new `test` function and the soft-voting implementation.

clinicadl/clinicadl/classify/inference.py
```diff
@@ -112,216 +136,202 @@ def inference_from_model(caps_dir,
...
+            infered_classes, metrics = inference_from_model_generic(
                 caps_dir,
                 tsv_path,
                 full_model_path,
                 options)
         elif (options.mode == 'roi'):
-            infered_classes = inference_from_roi_model()
+            infered_classes, metrics = inference_from_model_generic()
         else:
             print("Inference for this image mode is not implemented")
```
14thibea: If all modes use the same method we can remove this if else statement.

clinicadl/clinicadl/classify/inference.py
```diff
@@ -65,6 +73,9 @@ def classify(caps_dir, tsv_path, model_path, output_dir=None, gpu=True, prepare_
         tsv_path,
         model_path,
         json_file,
+        prefix_output,
+        output_dir,
+        no_labels,
```
14thibea, line 78: This option is not used yet, but its implementation can be done in another PR later.

clinicadl/clinicadl/classify/inference.py
```diff
@@ -112,216 +136,202 @@ def inference_from_model(caps_dir,
...
     # loop depending the number of folds found in the model folder
-    for fold_number in currentDirectory.glob(currentPattern):
-        model_fold_path = join(model_path, fold_number)
-        full_model_path = join(model_fold_path, 'models', best_model['best_acc'])
-        if not exists(join(full_model_path, 'model_best.pth.tar')):
+    for fold_dir in currentDirectory.glob(currentPattern):
+        split = int(str(fold_dir)[-1])
+        fold_path = join(model_path, fold_dir)
+        models_path = join(fold_path, 'models')
+        full_model_path = join(models_path, best_model['best_acc'])
```
14thibea, line 154: Instead of only evaluating 'best_acc' it could be possible to let the choice to the user, or by default to evaluate all the models selected on any metric.
This could be done in another PR.

clinicadl/clinicadl/cli.py
```diff
@@ -1010,26 +1014,30 @@ def parse_command_line():
...
-        help='Path to the folder where the model and the json file are stored.',
+        help='''Path to the folder where the model is stored. Folder structure
+                should be the same obtained during the training.''',
         default=None)
     classify_parser.add_argument(
-        'output_directory',
-        help='Folder containing results of the training.',
+        '-pre', '--prefix_output',
+        help='Prefix to name the files resulting from the classify task.',
+        type=str, default='prefix_DB')
```
14thibea, line 1023: I think the prefix option could be mandatory, then the user will better understand better which files have been written because they have to give a name.

clinicadl/clinicadl/cli.py
```diff
@@ -1010,26 +1014,30 @@ def parse_command_line():
...
+        '-nl', '--no_labels', action='store_true',
+        help='Add this flag if your dataset does not contain a ground truth.',
         default=False)
     classify_parser.add_argument(
-        '--use_extracted_slices',
-        help='''If True the extract slices are used, otherwise the whole
-                 MRI is loaded.''',
+        '--use_extracted_features',
+        help='''If True the extract slices or patche are used, otherwise the they
+                will be extracted on the fly (if necessary).''',
```
14thibea, line 1035: Maybe it could also be renamed this way in the `train` instead of having different names for all pipelines ?

14thibea (approved):
I added some comments that could be implemented later in another PR (next week I will have time on the project !)

clinicadl/clinicadl/cli.py
mdiazmel: Ok, I agree !

clinicadl/clinicadl/classify/inference.py
mdiazmel: Good point ! It was supposed to be removed, but I forgot. I will do it before merging.

PR #37, opened 2020-06-12
Improve tests

14thibea:
Tests were improved to detect any Exception raised during the training process.
One bug was fixed accordingly.


mdiazmel:
Hi @14thibea ,

Everything seems Ok. I'll merge it.

Thanks!

PR #39, opened 2020-06-16
Fixes softvoting

### LINKED ISSUE #33: Bug in soft voting for very small number of subjects
The test pipeline `test_train_cnn` occasionally crash with `train_slice` because there it may fail if one slice is always badly classified for all subjects. This happens when the number of subjects is very low (so hopefully, only in tests).
This could be fixed.

14thibea:
This PR closes https://github.com/aramis-lab/AD-DL/issues/33 in commit https://github.com/aramis-lab/AD-DL/commit/8fffdb6114e4a0edbef4e8a6b516773938bf4c79.
It also slightly optimizes and cleans the code in https://github.com/aramis-lab/AD-DL/commit/d4fabea556e3b7b21ac033030c1797286fd949ed.

It was tested on slice results published on Zenodo to check if the same results are obtained.

Now that this bug is fixed it will be possible to implement tests in the continuous integration !

### DIFF
--- a/clinicadl/clinicadl/image_level/utils.py
+++ b/clinicadl/clinicadl/image_level/utils.py
@@ -275,8 +275,8 @@ def test(model, dataloader, use_cuda, criterion, full_return=False):
 
             # Generate detailed DataFrame
             for idx, sub in enumerate(data['participant_id']):
-                row = [sub, data['session_id'][idx], labels[idx].item(), predicted[idx].item()]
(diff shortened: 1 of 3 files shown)


mdiazmel:
Thanks @14thibea ,
I'll merge this PR.
Just curious about how do you tested the published models? Do you use the `evaluation.py` script inside the corresponding folder?

14thibea:
Actually I only tested the method `soft_voting` and loaded myself the tsv files in which the results are stored in a ipython environment.

PR #41, opened 2020-06-16
Change behaviour of split variable

### LINKED ISSUE #25: Change default behavior for `split=None` in training of subject_level
For patch and slice, if `split = None` all the splits of the k-fold CV are done. It could be implemented for subject level and default value for `split` could be set to `None`.
Suggested in PR #20.

14thibea:
This PR closes https://github.com/aramis-lab/AD-DL/issues/25.

`split` is now a list of folds that are executed (default behaviour, all splits are trained).
As one run may correspond to several folds, the json file was moved at the root.

### DIFF
--- a/clinicadl/clinicadl/image_level/utils.py
+++ b/clinicadl/clinicadl/image_level/utils.py
@@ -16,7 +16,7 @@ from clinicadl.tools.deep_learning import EarlyStopping, save_checkpoint
 # CNN train / test  #
 #####################
 
-def train(model, train_loader, valid_loader, criterion, optimizer, resume, split, options):
+def train(model, train_loader, valid_loader, criterion, optimizer, resume, log_dir, model_dir, options):
     """
     Function used to train a CNN.
     The best model and checkpoint will be found in the 'best_model_dir' of options.output_dir.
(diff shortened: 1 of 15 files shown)


clinicadl/clinicadl/image_level/utils.py
```diff
@@ -16,7 +16,7 @@
 # CNN train / test  #
 #####################
 
-def train(model, train_loader, valid_loader, criterion, optimizer, resume, options):
+def train(model, train_loader, valid_loader, criterion, optimizer, resume, split, options):
```
mdiazmel: Why `split` is an additional parameter of this function? This information is not embedded in the `options` data structure?
I thing that we should keep the "training" functions with few arguments as possible and let the user of the function to manage the naming of the folders, files.

mdiazmel (commented):
Hi @14thibea,
The `split` parameter is now mandatory for train function on `image_level` (see below). Is this necessary? I wonder if the naming of  the folders/files can be done outside of the training functions in order to keep these functions as simple as possible.

clinicadl/clinicadl/image_level/utils.py
14thibea: Actually options.split contains the list of splits that must be done in the training, and train execute the computations for one particular split (so the arguments are different).
I can detach the naming of the folders / writers etc from the function so it will become independent from the split as you suggested !

14thibea: Done in https://github.com/aramis-lab/AD-DL/pull/41/commits/1b5848855f1b1bbe65fcf7a9746a0fb7121bf59a
But now there are more arguments in train...

mdiazmel: Yes, I see.... but I thing that this is better for visibility/control of files/folder created during the training.

mdiazmel:
It seems that this PR is ready to merge? Can you confirm it, @14thibea ?

14thibea:
It's ok, the tests all passed !

mdiazmel:
I'll merge it!

PR #46, opened 2020-06-19
Rename patch/slice files from extract task

mdiazmel:
Rename files produced by `extract patch` and `extract slice` in order to be coherent with naming in Clinica deeplearning-prepare-data pipeline.


mdiazmel:
Quick merge to assure training can handle new file names for patches and slices!

Issue #32, opened 2020-06-11
Fix generate pipelines


14thibea:
Paths of the images extracted by the preprocessing or extract have changed, but `generate` was not updated.
It will be fixed after the preprocessing pipelines will be finished.

mdiazmel:
There is another question linked to this issue: should we keep the preprocessing pipelines inside clinicadl or we should wrap preprocessing pipelines from Clinica?

PR #50, opened 2020-06-22
Renaming files in generate task

mdiazmel:
This PR renames files created using the task `generate` to use the same name convention than `clinicadl preprocessing`. It closes issue #32 .


14thibea:
Thank you for the PR !
I just tested and maybe the atlases that are stored at https://aramislab.paris.inria.fr/files/data/masks were not updated as I got masks of size (121, 145, 121) for linear preprocessing so it didn't match.

clinicadl/clinicadl/tools/data/generate_data.py
```diff
@@ -6,34 +6,53 @@
...
+from clinicadl.tools.inputs.filename_types import FILENAME_TYPE
+from clinicadl.tools.inputs.input import fetch_file
+from clinicadl.tools.inputs.input import RemoteFileStructure
+import tarfile
 
 
-def generate_random_dataset(caps_dir, tsv_path, output_dir, n_subjects, mean=0, sigma=0.5,
-                            preprocessing="linear", output_size=None):
+def generate_random_dataset(caps_dir, tsv_path, output_dir, n_subjects, mean=0,
+                            sigma=0.5, preprocessing="linear", output_size=None):
```
14thibea: If `generate` can be performed on different preprocessing that were conducted by the user, maybe the output_size argument used to up/downsampled the image is useless and can be removed.

clinicadl/clinicadl/tools/data/generate_data.py
```diff
@@ -63,40 +82,76 @@ def generate_random_dataset(caps_dir, tsv_path, output_dir, n_subjects, mean=0,
...
+        noisy_image_nii_path = join(output_dir, 'subjects', participant_id, 'ses-M00', 't1_linear')
+        noisy_image_nii_filename = participant_id + '_ses-M00' + FILENAME_TYPE['cropped'] + '.nii.gz'
+        if not exists(noisy_image_nii_path):
+            makedirs(noisy_image_nii_path)
+        nib.save(noisy_image_nii, join(noisy_image_nii_path, noisy_image_nii_filename))
 
 
 def generate_trivial_dataset(caps_dir, tsv_path, output_dir, n_subjects, preprocessing="linear",
                              mask_path=None, atrophy_percent=60, output_size=None, group=None):
     """
```
14thibea: This function was adapted to work with dartel outputs (linked to the `group` argument). Maybe we don't want to keep that as this preprocessing is never used for CNN training.
Same comment than before for `output_size`.

clinicadl/clinicadl/tools/data/generate_data.py
```diff
@@ -63,40 +82,76 @@ def generate_random_dataset(caps_dir, tsv_path, output_dir, n_subjects, mean=0,
...
+                tar_file.extractall(path_to_masks)
+                tar_file.close()
+                mask_path = join(path_to_masks, 'AAL2')
+            except RuntimeError:
+                print('Unable to extract donwloaded files')
+        except IOError as err:
+            print('Unable to download required templates:', err)
+            raise ValueError('''Unable to download masks, please donwload them
+                              manually at https://aramislab.paris.inria.fr/files/data/masks/
+                              and provide a valid path.''')
```
14thibea: I got this error, it said that the tar file was corrupted when dowloaded automatically... Then I downloaded the masks manually and gave the path to these masks and everything went well !

clinicadl/clinicadl/tools/data/utils.py
```diff
@@ -1,15 +1,15 @@
...
     if preprocessing == "linear":
         image_path = path.join(caps_dir, 'subjects', participant_id, session_id,
-                               't1', 'preprocessing_dl',
+                               't1_linear',
                                participant_id + '_' + session_id +
-                               '_space-MNI_res-1x1x1.nii.gz')
+                               FILENAME_TYPE['cropped'] + '.nii.gz')
     elif preprocessing == "extensive":
         image_path = path.join(caps_dir, 'subjects', participant_id, session_id,
                                't1', 'spm', 'segmentation', 'normalized_space',
```
14thibea: The preprocessing could be renamed `t1-linear` and `t1-extensive`. Also is it possible to use both of them or is only `t1-linear` supported now ?
Maybe we can implement `t1-extensive` later as the paths / filenames are not completely defined.

14thibea (approved):
I suggest some additional changes on arguments which became useless and could be removed (`output_size`, `group`) and other whose values could be renamed (`preprocessing`).

clinicadl/clinicadl/tools/data/utils.py
14thibea: Maybe we can reuse MRIDataset here to load the image based on the caps_dir and preprocessing and avoid to recode here the paths.

clinicadl/clinicadl/tools/data/generate_data.py
mdiazmel: Sorry, i forgot to change the hash for the new file ! It was done in the last commit (ab6b0e4).

14thibea: Yes it is working now !

clinicadl/clinicadl/tools/data/utils.py
mdiazmel: Only `t1-linear` is supported now. As you said, we need to define paths/filenames for `t1-expensive`. I renamed the options.

clinicadl/clinicadl/tools/data/generate_data.py
mdiazmel: Done!

clinicadl/clinicadl/tools/data/generate_data.py
mdiazmel: Done!

mdiazmel:
Hi @14thibea , 
I addressed all your suggestions. 
It seems ready to merge, are you agree?

14thibea (approved):

14thibea (commented):
I agree to merge !

PR #67, opened 2020-07-06
Fix bug in `add_demographics`

14thibea:
This fixes a bug in `add_demographics` method.


mdiazmel (commented):
It seems Ok. I'll merge!

PR #70, opened 2020-07-08
Add more complete env file

### LINKED ISSUE #61: Improve environment.txt generated by clinicadl train
`clinicadl train` outputs a txt file in which the version of some dependencies are stored. It could be improved so it could be used in a template for issues on `clinicadl train`.

14thibea:
This PR closes #61.

Right now the whole env versions is written using `pip freeze` but it might be a bit intrusive. Do you think we should restrict this file to the dependencies of clinicadl only ?


mdiazmel (commented):
It seems Ok. I'll merge this PR.

PR #71, opened 2020-07-09
Test transfer learning

### LINKED ISSUE #63: Transfer learning AE --> AE
This pre-existing feature was lost during the previous refactoring.
It could be interesting to put it back as AE training is costly and this feature was used during architecture search to transfer weights between two AE of different sizes.

14thibea:
This PR closes #63 and partly answers #37.

Automatic tests are coded to check transfer learning feasibility, moreover the feature that allows to initialize AE --> X with AE and X having a convolutional part of different size was restaured.
One larger model, Conv6_FC3 was added to test this feature.


14thibea:
Verbose of transfer learning will be handled in this PR

PR #68, opened 2020-07-07
Verbose

14thibea:
This PR resolves https://github.com/aramis-lab/AD-DL/issues/62.
The goal of this PR is to print messages with logging:
- messages with level info / debug or less are directed to stdout,
- other messages are directed to stderr. 

Specific loggers are created depending on the function and formatted the message to also print the date + the name of the logger.

Issue #60, opened 2020-07-06
Implement resume


14thibea:
It would be great to be able to resume a job that crashed and launch the remaining folds of a job (if only a subset was asked first).

PR #74, opened 2020-07-09
Add resume feature

14thibea:
This PR partly answers #60.

It implements the ability to resume crashed experiments by reloading the checkpoint and optimizer + training remaining folds.
Verbose will be added when #68 will be merged.


14thibea:
A .ended file is added at the fold level to keep track of job's end.
At the moment it is not homogeneous between multi and single/AE:
- single/AE outputs an empty .ended at the very end of the job
- multi outputs an empty.ended at the beginning of the job. Then each time a cnn ends its training its index is added to .ended. At the end the line "final evaluation" is added if the softvoting ended.

single/AE could also output the file at the beginning of the process (in main, near commandline writing) and add "final evaluation" if the final evaluation was performed.

PR #87, opened 2020-07-27
Remove no_labels from options

14thibea:
This PR removes the options `no_labels` from `classify`.
Indeed in this version the option was not implemented hence there is no effect when adding this flag (leaving in the command line is misleading).


mdiazmel (approved):
I'll merge this PR.

PR #89, opened 2020-07-27
Add warning if only one class is detected

### LINKED ISSUE #86: raise exception if the diagnoses all have the same label value
Diagnoses are pre-associated with class values
- 'CN': 0,
- 'AD': 1,
- 'sMCI': 0,
- 'pMCI': 1,
- 'MCI': 1

If someone wants to perform the binary classification CN vs sMCI, this could be a problem as the two diagnoses have the same class value --> raise an Exception to prevent that case.

14thibea:
The management of labels may not be easy to understand for newcomers, so a warning was added in case they perform a classification on classes with same label.
This closes #86.


mdiazmel (commented):
I think is ready to merge!

PR #103, opened 2020-09-24
Replace extract task by CLI from Clinica

### LINKED ISSUE #99: Wrap `clinica run deep-learning-prepare-data`
`clinicadl extract` should perform the same operation than `clinica run deep-learning-prepare-data` with the same arguments (currently this is not the case).
Then `clinicadl extract` should call `clinica run deep-learning-prepare-data` to prevent unwanted differences.

alexandreroutier:
Resolves #99


mdiazmel (commented):
Thanks @alexandreroutier,

I'll merge this PR.

PR #110, opened 2020-10-18
Use exist_ok=True option for makedirs

alexandreroutier:
This PR simplifies a little bit how folder are created in ClinicaDL.


mdiazmel (approved):
Thanks @alexandreroutier,
Well seen! 
I'll merge it.

PR #118, opened 2021-01-15
Generate fix

14thibea:
Fix minor bug in generate (will allow to add new notebook to online course)

### DIFF
--- a/clinicadl/clinicadl/tools/data/generate_data.py
+++ b/clinicadl/clinicadl/tools/data/generate_data.py
@@ -43,8 +43,6 @@ def generate_random_dataset(caps_dir, output_dir, n_subjects, tsv_path=None, mea
         A folder written on the output_dir location (in CAPS format), also a
         tsv file describing this output
 
-    Raises:
-
     """
     commandline_to_json({
         "output_dir": output_dir,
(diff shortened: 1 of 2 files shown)


clinicadl/clinicadl/tools/data/generate_data.py
```diff
@@ -150,7 +150,7 @@ def generate_trivial_dataset(caps_dir, output_dir, n_subjects, tsv_path=None, pr
 
     if n_subjects > len(data_df):
         raise ValueError("The number of subjects %i cannot be higher than the number of subjects in the baseline "
```
mdiazmel: Let's change this to f-strings:

```python
        raise ValueError(f"The number of subjects {n_subject} cannot be higher than the number of subjects in the baseline dataset of size {len(data_df)}"
```

mdiazmel (commented):
Hi @14thibea ,

Seems ok. 
A small detail, we try to adopt f-strings everywhere (at least for new code).

clinicadl/clinicadl/tools/data/generate_data.py
14thibea: Done in https://github.com/aramis-lab/AD-DL/pull/118/commits/d5d0d42575efbccefdd495fd5e99741bc50d6cce

PR #126, opened 2021-02-02
Add options cli

### LINKED ISSUE #123: t1-linear crash files cannot be written
Hi @mdiazmel,

There is a small bug in `t1-linear` pipeline when launched with ClinicaDL (it works with Clinica).
This is due to the last line of `t1_linear_cli.py`:
```
print_crash_files_and_exit(args.logname, pipeline.base_dir)
```

There is no `logname` in `args` in the case of ClinicaDL as this argument is added at a higher level in Clinica in `cmdline.py`:
```
    parser.add_argument("-l", "--logname",
                        dest='logname',
                        default="clinica.log",
                        metavar=('file.log'),
                        help='Define the log file name (default: clinica.log)')
```

This argument (and verbose) should be added to the general command line of `clinicadl preprocessing run`.

mdiazmel:
Add some options to the command line:

* `--verbose`
* `--version`
* `--logname`

This PR also make a little refactor of the command line and its (sub)parsers.

Closes #123

### DIFF
--- a/clinicadl/clinicadl/cli.py
+++ b/clinicadl/clinicadl/cli.py
@@ -250,7 +250,7 @@ def parse_command_line():
         dest='task',
         help='''****** Tasks proposed by clinicadl ******''')
 
-    #subparser.required = True
+    # subparser.required = True
 
     # Generate synthetic data
     generate_parser = subparser.add_parser(
@@ -265,7 +265,7 @@ def parse_command_line():
         dest='mode',
         help='''****** Synthetic datasets proposed by clinicadl ******''')
 
-    #generate_subparser.required = True
+    # generate_subparser.required = True
 
     # Positional arguments
     generate_rs_parent_parser = argparse.ArgumentParser(add_help=False)
--- a/clinicadl/clinicadl/cli.py
+++ b/clinicadl/clinicadl/cli.py
@@ -240,9 +240,6 @@ def parse_command_line():
                         action='store_true', default=False,
                         help="ClinicaDL's installed version")
 
-    parent_parser = argparse.ArgumentParser(add_help=False)
-    parent_parser.add_argument('--verbose', '-v', action='count', default=0)
-
     subparser = parser.add_subparsers(
         title='''Task to execute with clinicadl:''',
         description='''What kind of task do you want to use with clinicadl?
--- a/clinicadl/clinicadl/cli.py
+++ b/clinicadl/clinicadl/cli.py
@@ -566,7 +566,6 @@ def parse_command_line():
     # random search parsers
     rs_generate_parser = subparser.add_parser(
         'random-search',
-        parents=[parent_parser],
         help='Generate random networks to explore hyper parameters space.',
         formatter_class=argparse.ArgumentDefaultsHelpFormatter
     )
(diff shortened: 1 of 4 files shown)


clinicadl/clinicadl/cli.py
```diff
@@ -229,6 +229,17 @@ def parse_command_line():
...
+                        default="clinicaDL.log",
+                        metavar=('file.log'),
+                        help='Define the log file name (default: clinicaDL.log)')
+    parser.add_argument("-V", "--version",
+                        dest='version',
+                        action='store_true', default=False,
+                        help="ClinicaDL's installed version")
+
     parent_parser = argparse.ArgumentParser(add_help=False)
     parent_parser.add_argument('--verbose', '-v', action='count', default=0)
```
14thibea, line 243: Now that verbose is added for all parsers maybe we can remove this "parent parser"

clinicadl/clinicadl/cli.py
```diff
@@ -239,7 +250,7 @@ def parse_command_line():
         dest='task',
         help='''****** Tasks proposed by clinicadl ******''')
 
-    subparser.required = True
+    # subparser.required = True
```
14thibea, line 252: A naïve question: what is the difference when we comment this line?
I actually put this `required` argument to `True` everywhere

clinicadl/clinicadl/main.py
```diff
@@ -9,12 +9,18 @@ def main():
...
+        import clinicadl
+        print(f"ClinicaDL version is: {clinicadl.__version__}")
+        exit(0)
     if hasattr(args, 'use_cpu'):
         if not args.use_cpu and not torch.cuda.is_available():
             raise ValueError("No GPU is available. Please add the -cpu flag to run on CPU.")
 
-    args.func(args)
+    if not args.task:
+        parser.print_help()
```
14thibea, line 21: I think I understood the point ^^
But maybe we can remove more of them then?

14thibea (approved):
Seems ok to me!
I think we could just remove the parent parser to avoid a redundancy.

clinicadl/clinicadl/cli.py
mdiazmel: ~Done in 1a8e3ca~ ... Have to remove from other locations in the code...

clinicadl/clinicadl/cli.py
mdiazmel: The idea is to have a clear message and not an exception when launching the command line without any arguments!
In addition, it seems that required options are not advised ([see the note here](https://docs.python.org/3/library/argparse.html#required)).

clinicadl/clinicadl/main.py
mdiazmel: Yes, at some points they can be useful (I'm thinking in some of the subparsers).

clinicadl/clinicadl/cli.py
mdiazmel: Well, final solution (in a1674fa) is to keep the "parent parser" for the verbose option and a general purpose parser besides (for `--logname` and `--version`).

mdiazmel:
I'll merge this one!

PR #128, opened 2021-02-03
Simplify tsvtool - getlabels, split, analysis

14thibea:
The goal of this PR is to simplify some tsvtools.
Indeed the names of the columns vary depending on the dataset: for example the score MMSE can be named MMS, mmse, mmse_bl...
The user had to know what was the name of the column to launch the tool, now it will be found automatically.


mdiazmel (approved):
I'll merge this PR!

PR #131, opened 2021-02-08
[WIP] Add black and isort to homogeneize coding style

mdiazmel:
This PR implies a massive commit. All other PR should be closed/merged before proceeding. To remember: put a file with the commit ID in order to ignore revision if using `git blame` ([see here](https://black.readthedocs.io/en/stable/installation_and_usage.html#migrating-your-code-style-without-ruining-git-blame))


mdiazmel:
Seems that there are some incompatibilities between `black` and `pep8speaks` (note E203 and E304).

alexandreroutier:
Hi @mdiazmel,

It seems that `E203` is not compatible with `black`. However, i found clearer the spacing ` : ` instead of `: ` in the two cases (`clinicadl/clinicadl/quality_check/t1_linear/utils.py` and `clinicadl/clinicadl/tools/deep_learning/data.py`) so I would remove `E203` from PEP8 Speaks.

Regarding `E304`, I think you simply need to use `# fmt: off` before the `@pytest.fixture` instead of `fef cli_commands(request):` and the error should disappear.

Alex

alexandreroutier:
Hi again,

It is indeed explained here: https://black.readthedocs.io/en/stable/the_black_code_style.html ^^

Based on the documentation, `E203, E501, W503` would be the only rules to select and `W504` would be removed, same rationale some Clinica (3 rules instead of 14) 🤔

mdiazmel:
> Based on the documentation, E203, E501, W503 would be the only rules to select and W504 would be removed, same rationale some Clinica (3 rules instead of 14) 🤔

Thanks Alex for your suggestions!

I added those three to the `pep8speaks` setup. But it seems that it read the config file from the target branch and not from the fork. We'll see when the PR will be merged. For Clinica, the objective will be to arrive to the same 3 rules but it will be clearly harder!

14thibea:
Hi @mdiazmel 

Should we abandon this branch and redo it for v0.2.1 when the other PR of this version will be merged?

mdiazmel:
> Hi @mdiazmel
> 
> Should we abandon this branch and redo it for v0.2.1 when the other PR of this version will be merged?

Yes, sounds as the best option!

mdiazmel:
I close this one, to be opened later!

PR #150, opened 2021-04-15
Add new arguments for ROI in random-search

14thibea:
Fix random-search when chosen mode is ROI


mdiazmel (approved):
Seems ok!
I'll merge this fix, thanks!

PR #166, opened 2021-06-30
Maps manager

### LINKED ISSUE #144: Add customizable system of metrics
Current system only includes a limited number of metrics, it would be nice to have a template allowing to implement easiyl any metric.
Moreover, the selection criterion could be also chosen by the user (instead of always having balanced accuracy + loss)

### LINKED ISSUE #157: Multi-class classification
Provide a framework to perform multi-class classification as asked in https://groups.google.com/g/clinica-user/c/9LKl04t2wyY.

14thibea:
This PR implements the MapsManager, an interface able to write in a MAPS organization and get objects from it (trained models, predictions, metrics...)
The command line was a bit changed to adapt to modularity for some aspects (network, task, train / validation split loader).


docs/Train/Introduction.md
```diff
@@ -72,7 +72,7 @@ Options shared for all values of `mode` are organized in groups:
     computing the grey matter intensities.
 - **Cross-validation arguments**
     - `--n_splits` (int) is a number of splits k to load in the case of a k-fold cross-validation. Default will load a single-split.
-    - `--split` (list of int) is a subset of folds that will be used for training. By default all splits available are used. 
+    - `--folds` (list of int) is a subset of folds that will be used for training. By default all splits available are used.
```
omar-rifai, line 75: This is a bit confusing, you seem to be using 'split' and 'folds' interchangeably for an argument that is not the default k-fold algorithm behavior.
Idea: maybe we can just say '--train-partitions'?

14thibea: I agree, I have issues finding a good terms to define the number of splits / folds / partitions in a validation procedure (now we only have SingleSplit or KFold but it could be easily extended).
I would like to homogeneize everything with the same term but we have to choose which one before changing it everywhere.

PR #180, opened 2021-08-17
Fix split manager

14thibea:
Training is not possible for a list of folds > 0 without fold 0.
This PR fixes this bug.


mdiazmel:
Seems ok,

Can we merge this one?

14thibea:
Ok for me!

PR #183, opened 2021-08-17
Fix roi

14thibea:
ROI may crash if the user defined masks with 3D tensors.
Moreover the file system was different from the one of deeplearning-prepare-data!


mdiazmel (commented):
LGTM, I'll merge!

PR #187, opened 2021-08-19
Fix use of GPU on cluster

### LINKED ISSUE #177: Use pytorch.cuda instead of nvidia-smi
Currently the best GPU (selected based on memory available) is found with `nvidia-smi`
It could be better to use instead methods of `torch.cuda`

14thibea:
This PR detects if there is a restriction on the devices visible by cuda and lets cuda choose the gpu if it the case (else the gpu with the greatest amount of free memory is chosen).

Closes #177


mdiazmel:
Tested in an environment with multiple gpu devices (and `CUDA_VISIBLE_DEVICE` variable). It works fine!

PR #190, opened 2021-08-19
Added resume to command line

ravih18:
Small PR to add `resume` to command line. This pipeline was in the doc but not yet implemented with Click. This PR fix this.


mdiazmel:
I added a small test for the cli.

Thinking on adding tests for this feature (maybe a different PR). See issue #196 .

PR #191, opened 2021-08-19
Fix multi class

### LINKED ISSUE #182: Fix multi-class classification
Only binary classification is working as the task_manager is expecting only two outputs to write.

14thibea:
Closes #182


mdiazmel:
I'll merge this one!

PR #192, opened 2021-08-20
Fix random search

### LINKED ISSUE #185: Fix random search
Share readers between train and random search.
Use a TOML to define the configuration of the random search.

14thibea:
Closes #185 .
Also fixes and tests TOML configuration file readers.


clinicadl/clinicadl/train/train_cli.py
```diff
@@ -323,7 +323,7 @@ def cli(
         train_dict["unnormalize"] = not normalize
     if patience is not None:
         train_dict["patience"] = patience
-    if split != ():
+    if split:
```
ravih18, line 339: This can be removed since this option is a MapsManager.train parameters and is unused in paramters_dict.

clinicadl/clinicadl/train/train_utils.py
```diff
@@ -13,78 +13,62 @@ def get_train_dict(configuration_toml, preprocessing_json, task):
...
 
+    for command_name, code_name in renamed_dict:
+        train_dict[code_name] = train_dict.pop(command_name)
+
+    # GPU exception
+    train_dict["use_cpu"] = not train_dict.pop("use_gpu")
+
     # task dependent
     if task == "classification":
-        train_dict["loss"] = config_dict["Classification"]["optimization_metric"]
```
ravih18, line 65: Maybe put the task dependent before creating `train_dict` variable, and delete in the `config_dict` unused sections (regression and reconstruction) if task is classification for instance.

ravih18 (commented):
Thank you @14thibea for this PR. I leaved a comment that we can discuss to optimize a bit the train_dict creation.

clinicadl/clinicadl/train/train_utils.py
14thibea: Done in https://github.com/aramis-lab/clinicadl/pull/192/commits/5abf81f2258fce87a5258c228a0dddd1961bd4c2

clinicadl/clinicadl/train/train_cli.py
14thibea: I remove it from the dictionary at then end with `pop`, but I prefer to let the dictionary overwrite as `split` can be given in the TOML config file (or maybe we should remove it from the config file and only give it in command line?)

mdiazmel:
Seems ready to merge!

I'll do it, to reduce processing time during the tests!

PR #198, opened 2021-08-20
Extract

ravih18:
This PR propose two main improvement:

- Swap `use_extracted_features` from train to extract. Now the name in `save_features`.
- Change the command line: now the mode is a sub command to split options with the correct mode.

If the changes are validated I will update the documentation and tests.


mdiazmel:
Seems ready to merge!

PR #201, opened 2021-08-23
Fix extract

### LINKED ISSUE #199: Missing `discarded_slices` option in `clinicadl extract`
`discarded_slices` is not taken into account, leading to the extraction of all slices even if some needs to be discarded.

14thibea:
Closes #199.
Moreover slice extraction had some other issues:
- name ended with .nii.gz and not .pt
- with mode == `single` the number of channels was still 3


mdiazmel (approved):
LGTM!

PR #203, opened 2021-08-23
Rh testgpu

ravih18:
In this PR I propose a small function to test if GPU is available at the beginning of the pipeline. This test was previously done in the `clinicadl/clinicadl/main.py` file. But since the introduction of Click this file is not used and as been removed in this PR.


14thibea (approved):
Sounds good to me!

mdiazmel (approved):
Seems ok, 

I'll merge!

PR #210, opened 2021-08-30
Fix `clinicadl generate` module in last release

### LINKED ISSUE #208: No module named 'clinicadl.generate'
I followed every step from https://clinicadl.readthedocs.io/en/stable/Installation/ . When `clinicadl -h` is typed, I get the error: 
```
  File "/home/jajay/anaconda3/envs/clinicadlEnv/bin/clinicadl", line 5, in <module>
    from clinicadl.cmdline import cli
  File "/home/jajay/anaconda3/envs/clinicadlEnv/lib/python3.7/site-packages/clinicadl/cmdline.py", line 6, in <module>
    from clinicadl.generate.generate_cli import cli as generate_cli
ModuleNotFoundError: No module named 'clinicadl.generate'

```

14thibea:
It seems that generate is not seen as a package (see #208 ).
This PR should fix this issue.
(I suggest to merge it to master and release quickly as this blocks users).

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -24,7 +24,7 @@ pipeline {
              conda activate clinicadl_test
              echo "Install clinicadl using pip..."
              cd $WORKSPACE
-             pip install -e .
+             pip install  .
              pip install -r ./requirements-dev.txt
              # Show clinicadl help message
              echo "Display clinicadl help message"


Jenkinsfile
```diff
@@ -24,7 +24,7 @@ pipeline {
              conda activate clinicadl_test
              echo "Install clinicadl using pip..."
              cd $WORKSPACE
-             pip install -e .
+             pip install  .
```
mdiazmel, line 27: Not sure if this change anything (when running the tests)!

mdiazmel (commented):
LGTM!

I thing we should merge quickly and release a new version with the fix!

Jenkinsfile
14thibea: Not sure either, but locally for me it made the difference: it didn't work without `-e` when `__init__.py` was not here, whereas it worked with `-e`.

mdiazmel: Ok, I suppose that it works in editable mode because the folder isn't moved to different location, and it's able to detect the modules in the "code source". In any case, adding the `__init__.py` solves the problem and testing the package in real conditions seems a better idea.  I'll merge!

PR #204, opened 2021-08-24
Homogeneize `extract` with CapsDataset

14thibea:
This PR closes #184 and closes #43 .
Tools of mode extraction from extract were mutualized with CapsDataset.
It should be now possible to run train on `pet-linear` and `custom` preprocessing.

PR #212, opened 2021-09-01
Fix shepp logan

14thibea:
Shepp logan cannot be used as clinicadl train now need a JSON configuration file for preprocessing.
This PR fixes this problem.
(To be merged after #204 as it comes from this branch).


mdiazmel (approved):
I'll merge this PR!

PR #214, opened 2021-09-02
Fix train cli

ravih18:
Small fix on train cli:
- Check GPU was done before final value assignment
- In the test `if n_proc`, if n_proc is 0 then it doesn't work which cause some issue with Pytorch [W ParallelNative.cpp:206].


mdiazmel (approved):
LGTM!

I'll merge.

PR #218, opened 2021-09-07
Rh doc fix

ravih18:
Small fix in doc and docstring : 
- replace `clinica iotools missing_mods` with new name `clinica iotools check-missing-modalities`.
- replace `restriction_path` with option name `restriction_tsv` in doc.


mdiazmel:
@ravih18 a small conflict to solve before merging... due to recent changes introduced in `dev`.

PR #226, opened 2021-09-13
Fix multi_cohort

### LINKED ISSUE #225: multi-cohort is not working anymore
It is not possible to read a preprocessing.json file in a multi-cohort setting.
Moreover we don't know in which cohort this JSON file should exist...

14thibea:
Closes #225 .
A choice was made: the preprocessing.json file must exist in one of the CAPS only.
If preprocessing.json files with same name and different contents exist, this will be a problem for the user.


mdiazmel (approved):
Seem ready to merge!

PR #227, opened 2021-09-17
New command to get the informations on models

ravih18:
I propose here a new pipeline to print the list of different models (architecture) available in clinicadl.

There is an option to get the layers for a specific architecture.


14thibea:
When models are listed it should be nice to know to what task they are applicable (classification / regression OR reconstruction? 2D OR 3D?)

PR #248, opened 2021-10-26
Extract name

### LINKED ISSUE #242: Possible naming of JSON file produced by `extract`
This could help users to know their JSON files.
Moreover this would be necessary for users to perform the tests with pytest as described in the documentation.

14thibea:
Closes #242.
When the user choses the name of the JSON file, it may not follow the convention `extract_XXX.json` anymore.
If the user forgets to add the JSON extension on writing or reading, it is added automatically by ClinicaDL.

### DIFF
--- a/clinicadl/extract/extract_utils.py
+++ b/clinicadl/extract/extract_utils.py
@@ -46,14 +46,18 @@ def get_parameters_dict(
         parameters["acq_label"] = acq_label
         parameters["suvr_reference_region"] = suvr_reference_region
 
-    if json_name is None:
-        parameters["json_name"] = f"extract_{int(time())}.json"
-    elif not json_name.endswith(".json"):
-        parameters["json_name"] = f"{json_name}.json"
-    else:
-        parameters["json_name"] = json_name
+    parameters["json_name"] = compute_json_name(json_name)
 
     return parameters
+
+
+def compute_json_name(json_name: str) -> str:
+    if json_name is None:
+        return f"extract_{int(time())}.json"
+    elif not json_name.endswith(".json"):
+        return f"{json_name}.json"
+    else:
+        return json_name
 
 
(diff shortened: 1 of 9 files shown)


ravih18:
Hi, 

I just have a comment about the name of the option / variable `json_name`. I think it is better to keep the convention we use with the extension at the end of the option name: `xxxxx_json` instead of `json_name`. It could be `extract_json`, `preprocessing_json` (like in train command) or `output_json` for instance.

Otherwise it looks good to me.

clinicadl/extract/extract_utils.py
```diff
@@ -43,9 +46,20 @@ def get_parameters_dict(
         parameters["acq_label"] = acq_label
         parameters["suvr_reference_region"] = suvr_reference_region
 
+    parameters["json_name"] = compute_json_name(json_name)
+
     return parameters
 
 
+def compute_json_name(json_name: str) -> str:
+    if json_name is None:
```
ravih18: I am not sure, but didn't we decide to change our None test by `if not ..` instead of `if .. is None` ?

omar-rifai: Just for the records, we had to backtrack on some of these changes in Clinica. Particularly for empty pandas' Dataframes which crash on the check `if not <dataframe>`. For `str` it is indeed more idiomatic.

14thibea: We changed it in the argument detection, but not because of the style: it allowed to include in the test empty lists and tuples (with None values).
I didn't think it was a convention to follow everywhere

14thibea:
Thank you for the review :)
I renamed the variable `extract_json` it's better!
And I added it in the documentation also.

mdiazmel (approved):
Looks good. Ready to merge!

PR #261, opened 2021-12-14
Rh VAE features

### LINKED ISSUE #223: Differentiate components of loss in `compute_outputs_and_loss`
Could be interesting for VAE to differentiate the KL-loss and image reconstruction

### LINKED ISSUE #238: Parallelize extract
Add parallelization option for extract pipeline.
The easier solution is to reintegrate Nipype, but it can be done without Nipype too.
This closes issue #186

### LINKED ISSUE #249: Save reconstruction output in nifti format
It would be good to have the choice of the format in which we want to save save-tensors outputs (.pt or .nii.gz). Indeed nifti is better to visualize and save space.
Then we also would have to change the name of the command line ?

ravih18:
Here is the PR of my work during the sprint.

I propose some updates about VAE:
* New models (3D and 2D with both dense and spacial latent space),
* New metrics for evaluation of reconstruction,
* New dictionary for loss (returning all the different component of the loss),
* New options for VAE parameters (to change the architecture of the convolutional VAE: latent space, feature space, number of convolutional layers...).

Moreover I added the option to choose the number of workers in `extract` pipeline using the joblib library. 

There is also a new option in `save-tensor` to save the output as nifti file. This as not been tested yet (especially I am not sure about the affine matrix).

I finally think it is not necessary to add tests for VAE since reconstruction is already tested through autoencoders. Ad we don't have tests for ClinicaDL's models.

### DIFF
--- a/clinicadl/save_tensor/save_tensor.py
+++ b/clinicadl/save_tensor/save_tensor.py
@@ -14,9 +14,9 @@ def save_tensor(
     nifti=False,
     overwrite=False,
 ):
-    """TO CHANGE
-    This function loads a MAPS and predicts the global metrics and individual values
-    for all the models selected using a metric in selection_metrics.
+    """
+    This function loads a MAPS and compute reconstruction outputs and will save them in the MAPS
+    for all the models selected.
 
     Args:
         maps_dir (str): file with the model (pth format).
@@ -24,12 +24,12 @@ def save_tensor(
         caps_directory (str): path to the CAPS folder. For more information please refer to
             [clinica documentation](https://aramislab.paris.inria.fr/clinica/docs/public/latest/CAPS/Introduction/).
         tsv_path (str): path to a TSV file containing the list of participants and sessions to interpret.
-        labels: by default is True. If False no metrics tsv files will be written.
         gpu: if true, it uses gpu.
         selection_metrics: list of metrics to find best models to be evaluated.
         diagnoses: list of diagnoses to be tested if tsv_path is a folder.
         verbose: level of verbosity.
         multi_cohort (bool): If True caps_directory is the path to a TSV file linking cohort names and paths.
+        nifti (bool): If True will save the outputs as nifti files instead of Pytorch tensors.
         overwrite (bool): If True former definition of data group is erased
     """
     verbose_list = ["warning", "info", "debug"]
--- a/clinicadl/save_tensor/save_tensor.py
+++ b/clinicadl/save_tensor/save_tensor.py
@@ -8,12 +8,10 @@ def save_tensor(
     caps_directory,
     tsv_path,
     gpu=True,
-    n_proc=0,
-    batch_size=1,
-    # prepare_dl=True,
     selection_metrics=None,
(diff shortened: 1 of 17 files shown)


mdiazmel:
Cool! 

If I understand well, you also introduced `joblib` to manage parallelization when launching `clinicadl extract`, isn't?

If that's the case, could you please update the description of the PR?

It seem's a huge PR with a lot of things inside. Let us know when it'll be ready for review, before it becomes "unmergeable" ;)

clinicadl/save_tensor/save_tensor.py
```diff
@@ -8,12 +8,10 @@ def save_tensor(
...
-    n_proc=0,
-    batch_size=1,
-    # prepare_dl=True,
     selection_metrics=None,
     diagnoses=None,
     multi_cohort=False,
+    nifti=False,
     overwrite=False,
 ):
     """TO CHANGE
```
mdiazmel: There's still something to change?  Otherwise this label can be removed. 
Edit: after reading the docstring for this function, I confess that it's not so clear. Could you edit it, please ?

clinicadl/utils/maps_manager/maps_manager.py
```diff
@@ -1032,6 +1055,70 @@ def _test_loader(
...
+        self,
+        dataset,
+        data_group,
+        split,
+        selection_metrics,
+        gpu=None,
+        network=None,
+    ):
+        """
+        Compute the output nifti image and saves them in the MAPS.
```
mdiazmel: No sure to understand why the nifty image is stored in the MAPS ... 
BTW:
```suggestion
        Computes the output nifti images and saves them in the MAPS.
```
or 
```suggestion
        Computes the output nifti image and saves it in the MAPS.
```

mdiazmel (approved):
HI @ravih18 ,

Looks good! Seems a nice idea to use `joblib` to management parallelization in extract. Maybe we can use it also for other _preprocessing_  tasks.

I added a couple of comments that need clarification on the docstrings. After that, I suppose that we can merge.

BTW, PR#215 can be closed without merging, isn't?

ravih18:
Hi @mdiazmel , 

Thanks for your review. 

Indeed parallelizing the preprocessing with `joblib` is a great idea. I opened the issue #280 for it.

In the current version, the `save-tensor` command will reconstruct the image as a `.pt` tensor which might not be convenient for the users (heavy, not easy to visualize, not easy to use in other programs...). This is why I added the option in the command to save the output images as nifti file that might solve some of these issues. As the tensors where saved in the MAPS, I did the same for the nifti file !

PR#215 can indeed be closes without merging !

mdiazmel:
> In the current version, the `save-tensor` command will reconstruct the image as a `.pt` tensor which might not be convenient for the users (heavy, not easy to visualize, not easy to use in other programs...). This is why I added the option in the command to save the output images as nifti file that might solve some of these issues. As the tensors where saved in the MAPS, I did the same for the nifti file !

Thanks for the explanation. 

I'll merge it!

PR #263, opened 2021-12-16
Changed ValueError exception to more appropriate exception

### LINKED ISSUE #175: Improve Error Management
Several errors of the same function are raised by `ValueError`. Different names should be defined and given.
Moreover Errors should not be mentioned in the docstring unless they are explicitely raised in the function.

ravih18:
In this PR I propose some new exceptions for errors raised in ClinicaDL (closes #175 ).

### DIFF
--- a/clinicadl/tsvtools/analysis/analysis.py
+++ b/clinicadl/tsvtools/analysis/analysis.py
@@ -129,7 +129,9 @@ def demographics_analysis(merged_tsv, formatted_data_path, results_path, diagnos
             elif sexM:
                 results_df.loc[diagnosis, "sexM"] += 1
             else:
-                raise ValueError(f"Patient {subject} has no sex")
+                raise ValueError(
+                    f"The field 'sex' for patient {subject} can not be determided"
+                )
 
             cdr = merged_df.at[(subject, first_session_id), fields_dict["CDR"]]
             if cdr == 0:
--- a/clinicadl/tsvtools/analysis/analysis.py
+++ b/clinicadl/tsvtools/analysis/analysis.py
@@ -129,7 +129,7 @@ def demographics_analysis(merged_tsv, formatted_data_path, results_path, diagnos
             elif sexM:
                 results_df.loc[diagnosis, "sexM"] += 1
             else:
-                raise ValueError("Patient %s has no sex" % subject)
+                raise ValueError(f"Patient {subject} has no sex")
 
             cdr = merged_df.at[(subject, first_session_id), fields_dict["CDR"]]
             if cdr == 0:
--- a/clinicadl/tsvtools/analysis/analysis.py
+++ b/clinicadl/tsvtools/analysis/analysis.py
@@ -130,7 +130,7 @@ def demographics_analysis(merged_tsv, formatted_data_path, results_path, diagnos
                 results_df.loc[diagnosis, "sexM"] += 1
             else:
                 raise ValueError(
-                    f"The field 'sex' for patient {subject} can not be determided"
+                    f"The field 'sex' for patient {subject} can not be determined"
                 )
 
             cdr = merged_df.at[(subject, first_session_id), fields_dict["CDR"]]
--- a/clinicadl/tsvtools/analysis/analysis.py
+++ b/clinicadl/tsvtools/analysis/analysis.py
@@ -75,8 +75,8 @@ def demographics_analysis(merged_tsv, formatted_data_path, results_path, diagnos
         diagnosis_path = path.join(formatted_data_path, diagnosis + ".tsv")
         if not path.exists(diagnosis_path):
             print(
-                "TSV file with all sessions was not found for diagnosis %s. "
-                "Loads baseline version instead." % diagnosis
+                f"TSV file with all sessions was not found for diagnosis {diagnosis}. "
+                f"Loads baseline version instead."
             )
             diagnosis_path = path.join(formatted_data_path, diagnosis + "_baseline.tsv")
         diagnosis_df = pd.read_csv(diagnosis_path, sep="\t")
--- a/clinicadl/tsvtools/getlabels/getlabels.py
+++ b/clinicadl/tsvtools/getlabels/getlabels.py
@@ -19,7 +19,7 @@ from typing import Dict, List
 import numpy as np
 import pandas as pd
 
-from clinicadl.utils.exceptions import ArgumentError, ClinicaDLTSVError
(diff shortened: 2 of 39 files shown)


clinicadl/tsvtools/analysis/analysis.py
```diff
@@ -129,7 +129,7 @@ def demographics_analysis(merged_tsv, formatted_data_path, results_path, diagnos
             elif sexM:
                 results_df.loc[diagnosis, "sexM"] += 1
             else:
-                raise ValueError("Patient %s has no sex" % subject)
+                raise ValueError(f"Patient {subject} has no sex")
```
mdiazmel: This phrase is kind of weird :)
I would say:
```suggestion
                raise ValueError(f"The field 'sex' for patient {subject} can not be determided")
```

clinicadl/utils/exceptions.py
```diff
@@ -0,0 +1,32 @@
+class DownloadError(Exception):
+    """Base class for download errors exceptions."""
+
+    pass
+
+
+class ArgumentError(ValueError):
```
mdiazmel: Sometimes `ArgumentError` is used from  `ctypes` (see file `clinicadl/generate/generate_utils.py`).
But here, `ArgumentError` is inheriting from `ValueError`. From what I known, `ArgumentError` from the module `ctypes` must be used when the function that raises the exception is an external function, isn't ? I thing that this point must be clarified.

clinicadl/utils/split_manager/split_manager.py
```diff
@@ -1,10 +1,13 @@
 import abc
+from ctypes import ArgumentError
```
mdiazmel: Is it the right `ArgumentError`? (cf. comment above).

mdiazmel (commented):
Hi @ravih18 ,

Thanks, nice work for better handling the exceptions. I posted a couple of comments/questions.

clinicadl/utils/exceptions.py
ravih18: I changed the `ArgumentError` I created to `ClinicaDLArgumentError` and changed to wrong import (from automatic import made by VScode).

mdiazmel (approved):

clinicadl/tsvtools/getlabels/getlabels.py
```diff
@@ -464,8 +465,8 @@ def get_labels(
             session = filename.split("_")[-1]
             missing_mods_df = pd.read_csv(path.join(missing_mods, file), sep="\t")
             if len(missing_mods_df) == 0:
-                raise ValueError(
-                    f"Empty DataFrame at path {path.join(missing_mods, file)}"
+                raise ClinicaDLTSVError(
+                    f"Given TSV file at {path.join(missing_mods, file)} load an empty DataFrame."
```
14thibea: load**s** instead of load?

clinicadl/utils/maps_manager/maps_manager.py
```diff
@@ -196,12 +202,6 @@ def predict(
             overwrite: If True erase the occurrences of data_group.
             label: Target label used for training (if network_task in [`regression`, `classification`]).
             label_code: dictionary linking the target values to a node number.
-
-        Raises:
-            ValueError:
-                when trying to overwrite train or validation data groups
-                when caps_directory or df are given but data group already exists
-                when caps_directory or df are not given and data group does not exist
```
14thibea, line 208: Should this be replaced by the new errors launched?

clinicadl/utils/maps_manager/maps_manager.py
```diff
@@ -352,12 +352,6 @@ def save_tensors(
             diagnoses (list[str]): List of diagnoses to load if tsv_path is a split_directory.
             gpu (bool): If given, a new value for the device of the model will be computed.
             overwrite (bool): If True erase the occurrences of data_group.
-
-        Raises:
-            ValueError:
-                when trying to overwrite train or validation data groups
-                when caps_directory or df are given but data group already exists
-                when caps_directory or df are not given and data group does not exist
```
14thibea, line 364: same as before

clinicadl/utils/maps_manager/maps_manager.py
```diff
@@ -470,12 +464,6 @@ def interpret(
             gpu (bool): If given, a new value for the device of the model will be computed.
             overwrite (bool): If True erase the occurrences of data_group.
             overwrite_name (bool): If True erase the occurrences of name.
-        Raises:
-            ValueError:
-                when trying to overwrite train or validation data groups
-                when caps_directory or df are given but data group already exists
-                when caps_directory or df are not given and data group does not exist
-                when name already exists and overwrite_name is False
```
14thibea, line 482: same as before

clinicadl/utils/maps_manager/maps_manager.py
```diff
@@ -1312,8 +1300,8 @@ def _check_data_group(
             label (str): label name if applicable
 
         Raises:
-            ValueError:
-                when trying to overwrite train or validation data groups
+            MAPSError when trying to overwrite train or validation data groups
+            ClinicaDLArgumentError:
                 when caps_directory or df are given but data group already exists
                 when caps_directory or df are not given and data group does not exist
```
14thibea, line 1312: Actually I think one of the goal of #175 was to avoid to have one error managing several cases. Maybe we could have two different errors for these cases suech ClinicaDLUndefinedGroup ClinicaDLOverDefinedGroup for example?

14thibea (approved):
Thank you for these changes! Many errors were significantly improved.
I left some comments but you may not address them if you don't think it is necessary.

clinicadl/utils/maps_manager/maps_manager.py
ravih18: This also one of the reflexion I had when adding the new errors message. However I think having more general name for error is more clear, thats is to say raison a `ClinicaDLArgumentError` for every errors related to command line arguments might be easier to understand for a new user. In many cases I tried to separate the error type, but in this case it is better to keep it like this (or raise the error in a sub function or at higher level).

clinicadl/utils/maps_manager/maps_manager.py
ravih18: It has been replaced by MAPSerror.

14thibea: What I meant is that it should be put in the docstring then

PR #241, opened 2021-10-21
Use JSON configuration files

14thibea:
Closes #194 and adresses partly #188.
It also implements the new commandline of `train` as proposed by @ravih18 in #246:

* clinicadl
    * train
        * classification
        * regression
        * reconstruction
        * any new tasks such as segmentation...
        * resume
        * from_json
        * list_models
        * any new training feature

The name `from_json` is not ideal, do not hesitate if you think of something better

To implement it a set of new files in `clinicadl/train/tasks` were added, and `clinicadl/train/train_cli.py` was substantially changed.
The list of common options of all train pipelines were implemented in `clinicadl/train/train_option.py`.
Moreover the way default options are handled evolved to be managed by the MapsManager (cf #244). This is why `maps_manager_utils.py` was added and `maps_manager.py` was changed.

Changes also impacted `random-search` as it shares a lot of arguments with `train`, and in particular the way default values are handled.

Finally, the implementation of `from_json` was a good opportunity to test determinism in ClinicaDL. As pytorch does not handle yet the determinism of max-pooling layers with GPU training, the model `Stride_Conv5_FC3` was added to perform the same task without pooling layers (instead it has a stride value of 2).

PR #268, opened 2021-12-20
Allow choice of optimizer and loss

### LINKED ISSUE #254: Add optimizer choice
Choose all optimizers in `torch.optim` with parameters `lr` and `weight_decay` to allow to use them with the commandline.
Right now, Adam is hard-coded.

14thibea:
To be merged after #241 

Currently the optimizer is hard-coded to `Adam` and the loss depends on the task performed.
It will be now possible to give other choices to the user.


mdiazmel (approved):
LGTM!

Can be merged!

PR #269, opened 2021-12-21
Fix quality check

14thibea:
Fix both quality check pipelines + add a test on quality check.
continuous integration data for `t1-volume` must still be added to the server.


mdiazmel (approved):
Data was already uploaded to the CI agents. Seems ready to merge!

PR #281, opened 2022-01-19
Rh doc

ravih18:
I changed `save-tensor` section name as it was not really clear (@nburgos ) from `Network output serialization` to `Save reconstruction outputs`.

Fixed also an error in the options of `tsvtool split`.


mdiazmel (approved):
LGTM!

PR #282, opened 2022-01-21
Fix toml default arguments reading

ravih18:
This PR fixes a bug introduced in the PR #241 (commit `17fcf21`). 

The `train_dict` constructor in `clinicadl/train/train_utils.py` didn't load default argument in the TOML file in `clinicadl/resources/config/train_config.toml`.

The fact that the CI didn't "detect" this error is a problem IMO.

### DIFF
--- a/clinicadl/random_search/random_search_utils.py
+++ b/clinicadl/random_search/random_search_utils.py
@@ -4,7 +4,7 @@ from typing import Any, Dict, Tuple
 
 import toml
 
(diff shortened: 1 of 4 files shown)


ravih18:
I also removed default value for `selection_metrics` in the command line option as it was overwriting the value given by users in the configuration files (and thus was only saving the best model according to the loss).

mdiazmel (approved):
Looks good! 
I'll merge it!

PR #294, opened 2022-02-09
Logger issue

ravih18:
I removed the default argument setting verbose to `info` in training.

However I think it would be better to handle logging as it is done in Clinica. This is a long work so I wont include it in this PR.


mdiazmel (approved):

mdiazmel:
> However I think it would be better to handle logging as it is done in Clinica. This is a long work so I wont include it in this PR.

Yes, I agree. For a next iteration!

PR #309, opened 2022-06-28
Update Python and other dependencies versions

mdiazmel:
Release constrains  Python versions (>=3.8, <3.11) and update Clinica version to >=0.7.0


ravih18 (approved):
LGTM

Issue #331, opened 2022-07-25
Fix list_models function


camillebrianceau:
The list_models function can't access some networks architecture (problem with the input size).

mdiazmel:
Closed by PR #332

PR #332, opened 2022-07-25
Fix list_models bug

camillebrianceau:
Fix issue #331 

Add a static method in the Network class (get_input_size). 
Now, the command "clinicadl train list_models -a  _architecture_ " displays the network architecture even if you don't precise the input_size in the command line.

### DIFF
--- a/clinicadl/utils/network/network.py
+++ b/clinicadl/utils/network/network.py
@@ -47,7 +47,11 @@ class Network(nn.Module):
 
     @staticmethod
     @abc.abstractmethod
-    def get_input_size():
+    def get_input_size() -> str:
+        """
+        This static method is used for list_models command.
+        Must return the shape of the input size expected (C@HxW or C@HxWxD) for each architecture.
+        """
         pass
 
     @abc.abstractproperty
--- a/clinicadl/utils/network/network.py
+++ b/clinicadl/utils/network/network.py
@@ -45,6 +45,11 @@ class Network(nn.Module):
                 free_gpu = argmax(memory_list)
                 return f"cuda:{free_gpu}"
 
+    @staticmethod
+    @abc.abstractmethod
+    def get_input_size():
+        pass
+
     @abc.abstractproperty
     def layers(self):
         pass
--- a/clinicadl/utils/network/vae/vae_utils.py
+++ b/clinicadl/utils/network/vae/vae_utils.py
@@ -18,7 +18,6 @@ class EncoderLayer2D(nn.Module):
         kernel_size=4,
         stride=2,
         padding=1,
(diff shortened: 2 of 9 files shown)


clinicadl/utils/network/network.py
```diff
@@ -45,6 +45,11 @@ def _select_device(gpu):
                 free_gpu = argmax(memory_list)
                 return f"cuda:{free_gpu}"
 
+    @staticmethod
+    @abc.abstractmethod
+    def get_input_size():
+        pass
```
ravih18, line 55: It would be great to add a doc string which specifies that this static method is used for `list_models` command, it is fixed for each architecture and the shape of the input size expected (C@HxW or C@HxWxD)

clinicadl/utils/network/vae/vanilla_vae.py
```diff
@@ -249,7 +261,7 @@ def __init__(
             # [0, 0, 0],
             # [0, 0, 1],
         ]
-
+        print(input_size)
```
ravih18: This print is not necessary.

clinicadl/utils/network/vae/vae_utils.py
```diff
@@ -18,7 +18,6 @@ def __init__(
         kernel_size=4,
         stride=2,
         padding=1,
-        output_padding=0,
```
ravih18, line 21: This might be usefull sometime, I am not sure we can remove it.

ravih18 (commented):
I made a few comments. Especially on the `clinicadl/utils/network/vae/vae_utils.py`, the `output_padding` variable might be usefull.

clinicadl/utils/network/vae/vae_utils.py
```diff
@@ -28,7 +28,6 @@ def __init__(
                 kernel_size,
                 stride=stride,
                 padding=padding,
-                output_padding=output_padding,
```
ravih18, line 31: If the argument output padding exists, it then needs to be passed here.

clinicadl/utils/network/vae/vae_utils.py
```diff
@@ -329,7 +328,6 @@ def __init__(
                 DecoderLayer2D(
                     last_layer_channels * 2 ** (i),
                     last_layer_channels * 2 ** (i - 1),
-                    output_padding=output_padding[i],
```
ravih18, line 332: same comment

ravih18 (approved):

mdiazmel (approved):
Seems ok.
There is a conflict with the `.pre-commit-config.yaml` file that avoids the merge. @camillebrianceau , could you solve this please?

Issue #273, opened 2021-12-27
Add properties to models


ravih18:
It would be great to add some properties to the `network` class such as `task` (list of tasks associated to the model), `is_3D` (bool) and adding a description in the docstring of each models available in ClinicaDL.
In that way we could improve the informations displayed in the `clinicadl train list_model` command line that would help a user (in which case should I use this model, for which task we estimate that a model in well performing, why did we add a model...).

ravih18:
An other good information that could be added is the reference paper from which the model has been implemented and in which paper it has been used.

ravih18:
Closed by #333

PR #333, opened 2022-07-27
Add properties to models 

camillebrianceau:
issue #273

Add 2 new properties (static method) in the Network class, get_dimension and get_task.
Add a little description in the docstring of each models available in ClinicaDL
Add an option to  `clinicadl train list_models`.

- If you choose a specific model with architecture, you will get all the basic information of the model. (docstring +properties)
- If you add the flag -model_layers, this pipeline will show the whole model layers.   
- If you add a specific shape with -input_size, it will show the whole model layers with your chosen input size.


clinicadl/train/list_models_cli.py
```diff
@@ -6,26 +6,35 @@
     "-a",
     "--architecture",
     type=str,
-    help="Architecture of the chosen model to display.",
+    help="Basic informations about the chosen architecture",
```
ravih18: I would not use 'Basic' here, and Information doesn't take an 's' in english. Here the help is more a description of the --architecture option, not really a description of what does the command do with this option.

clinicadl/train/list_models_cli.py
```diff
@@ -6,26 +6,35 @@
...
-    show_default=True,
     help="Size of the input image in the shape C@HxW if the image is 2D or C@DxHxW if the image is 3D.",
 )
+@click.option(
+    "-m",
+    "--model_layers",
+    type=bool,
+    default=False,
+    is_flag=True,
+    help="Architecture of the chosen model to display.",
```
ravih18: Same, here it would more be something like `Display the detailed Pytorch architecture`

clinicadl/train/list_models_cli.py
```diff
@@ -6,26 +6,35 @@
...
+)
 def cli(
     architecture,
     input_size,
+    model_layers,
 ):
     """Show the list of available models in ClinicaDL.
-    If you choose a specific model with architecture, this pipeline will show the whole model layers.
+    If you choose a specific model with architecture, you will get all the basic information of the model.
```
ravih18: IMO it is better to keep it impersonal: `the pipeline will show` or `information about how to use this model will be displayed` instead or `you ...`.

clinicadl/train/list_models_cli.py
```diff
@@ -6,26 +6,35 @@
...
+    If you add the flag -model_layers, this pipeline will show the whole model layers.
```
ravih18: the option name misses a dash -

clinicadl/train/list_models_cli.py
```diff
@@ -6,26 +6,35 @@
...
+    If you choose a specific shape with -input_size, it will show the whole model layers with your chosen input size.
```
ravih18: same

clinicadl/train/train_utils.py
```diff
@@ -89,10 +89,11 @@ def build_train_dict(config_file: str, task: str) -> Dict[str, Any]:
     return train_dict
 
 
-def get_model_list(architecture=None, input_size=(128, 128)):
+def get_model_list(architecture=None, input_size=None, model_layers=None):
```
ravih18: I think the default value for model_layer would rather be False than None

clinicadl/train/train_utils.py
```diff
@@ -89,10 +89,11 @@ def build_train_dict(config_file: str, task: str) -> Dict[str, Any]:
...
     """
     Print the list of models available in ClinicaDL.
-    If architecture is given, it prints the details of the specified model.
+    If architecture is given, it prints some informations of the specified model.
+    If you add the flag -model_layers, this pipeline will show the whole model layers.
```
ravih18: Same, options have two dashes --

clinicadl/train/train_utils.py
```diff
@@ -116,17 +117,62 @@ def get_model_list(architecture=None, input_size=(128, 128)):
...
+        if not input_size:
+            input_size = model_class.get_input_size()
+
+        secho(f"\nInformation for '{architecture}' network:", bold=True)
+
+        if not model_layers:
+
+            print(model_class.__doc__)
+            dimension = model_class.get_dimension()
+            if dimension == 0:
```
ravih18: I think this can be shorten by factorizing the print at the end on the if..elif

clinicadl/train/train_utils.py
```diff
@@ -116,17 +117,62 @@ def get_model_list(architecture=None, input_size=(128, 128)):
...
+                print(
+                    "\tThis model can only deal with".expandtabs(4),
```
ravih18: I am not sure to understand exactly what is the purpose of expandtabs

clinicadl/utils/network/network.py
```diff
@@ -45,6 +45,32 @@ def _select_device(gpu):
...
+        """This static method is used for list_models command.
+        Return 0 if the model is only 2D.
+        Return 1 if the model is only 3D.
+        Return 2 if the model can be both 2D and 3D.
+        """
+        pass
+
+    @staticmethod
+    @abc.abstractmethod
+    def get_task():
```
ravih18: We could maybe also fix the type of the output which is list.

ravih18 (commented):
I made a few comments on this PR.

Some general comment:
* I do not understand why two different ways of print bold text are used (secho and \033[1m string). It would be great to homogenize between print and echo.
* I think that if the --model layer option is passed, then the function should still print the informations.
* Some references are missing in dosctrings and should be added.
* Using 0, 1 or 2 for dimension encoding is not so clear.

ravih18:
Also, the doc needs to be updated.

clinicadl/train/train_utils.py
```diff
@@ -118,14 +118,50 @@ def get_model_list(architecture=None, input_size=None):
...
+        elif dimension == "2D or 3D":
+            shape_str = "C@HxW or C@DxHxW,"
+
+        shape_str = "\n\tThe input must be in the shape ".expandtabs(4) + shape_str
+        input_size_str = f" for example input_size can be {input_size}."
+
+        task_list = model_class.get_task()
+        task_str = f"\n\tThis model can be used for {task_list[0]}".expandtabs(4)
+        list_size = len(task_list)
+        if list_size > 0:
```
ravih18, line 140: This can be simplified by using the `join` method for string the following way: `" or ".join(task_list)`

ravih18 (commented):
Thanks for this PR, I just added a last comment that could simplify a bit the code.

mdiazmel (approved):
LGTM! 
Will be merged before release v.1.1.1

PR #339, opened 2022-08-09
Minor fixes + addition

14thibea:
Hello ClinicaDL team,

I recently worked with ClinicaDL on a projet and encountered some minor problems, so this PR intends to fix them:
- individual attribution maps were not correctly saved,
- NVML was not correctly installed on my machine and it raised an error that could be avoided,
- multi-cohort did not work with ROI caps dataset, as a choice could not be made between CAPS directories to choose the masks.

Moreover there are some minor additions:
- a new interpretability method Grad-CAM (with a test),
- the user can choose their own label_code (if for some reasons they want to group different classes under the same label).
- a getter in MapsManager to get the best epoch.

If this PR is too hard to read as such I can also split the commits in different branches.

Best,
Elina


ravih18 (commented):
LGTM, just need to add the new interpretation method (and the new options it implied) to the doc.

ravih18 (approved):

PR #342, opened 2022-09-08
Logger updates

### LINKED ISSUE #295: Improve logging management
Right now the logging is centralized in MAPS Manager even for commands that doesn't require MAPS Manager. 

It would be great to separate the different loggers of different part of the software as it is done in Clinica.

ravih18:
In this PR I propose a new logger for clinicadl with new handlers and new formatters.

Now the default logging mode is `INFO` and adding the `-v` verbose option sets the logger in `DEBUG`.

By default, the INFO and WARNING will be handled by stdout, by a minimal formatting (Hour:Min:Sec - Message). Then is the verbose option is passed, the logger will save the detailed logs in a `clinicadl_debug.log` file in the format `date time - name - level - message`.

I addition, as specified in Issue #295, I named the loggers of the different modules (maps_manager, task_manager...) with a specific name.


camillebrianceau (approved):

clinicadl/utils/maps_manager/__init__.py
```diff
@@ -1,8 +1,8 @@
+from ..logwriter import LogWriter
```
mdiazmel: Change to use absolute paths
```suggestion
from clinicadl.utils.logwriter import LogWriter
```

clinicadl/utils/logwriter.py
```diff
@@ -19,37 +19,57 @@ def filter(self, record):
...
+    info_formatter = Formatter("%(asctime)s - %(message)s", "%H:%M:%S")
+    debug_formatter = Formatter(
+        "%(asctime)s - %(name)s - %(levelname)s: %(message)s", "%Y-%m-%d %H:%M:%S"
+    )
+
+    fh = FileHandler("clinicadl_debug.log")
+    fh.setLevel(DEBUG)
+    fh.setFormatter(debug_formatter)
+
+    ch_info = StreamHandler(stdout)
```
mdiazmel: I would add this, to be have more autodocumented code:
```suggestion
    ch_info = StreamHandler(stream=sys.stdout)
```

clinicadl/utils/logwriter.py
```diff
@@ -19,37 +19,57 @@ def filter(self, record):
         return self.err
 
 
-def setup_logging(verbosity: int = 0) -> None:
+def setup_logging(verbose: bool = False) -> None:
```
mdiazmel: I would rather be in favor of putting this function inside `cmdline.py` and keep the class `LogWriter` and the file `logwriter.py` on the original location (this class only concerns the MAPS operations).

clinicadl/utils/logwriter.py
```diff
@@ -19,37 +19,57 @@ def filter(self, record):
...
+        getLogger,
+    )
     from sys import stderr, stdout
 
-    # Cap max verbosity level to 2.
-    verbosity = min(verbosity, 2)
+    logging_level = "DEBUG" if verbose else "INFO"
 
     # Define the module level logger.
     logger = getLogger("clinicadl")
```
mdiazmel: Call this `clinicadl_logger` to avoid confusion with a `root` logger.

clinicadl/utils/logwriter.py
```diff
@@ -19,37 +19,57 @@ def filter(self, record):
...
-        logger.handlers.clear()
+    warning_formatter = Formatter(
+        "%(asctime)s - %(levelname)s: %(message)s", "%H:%M:%S"
+    )
+    info_formatter = Formatter("%(asctime)s - %(message)s", "%H:%M:%S")
+    debug_formatter = Formatter(
+        "%(asctime)s - %(name)s - %(levelname)s: %(message)s", "%Y-%m-%d %H:%M:%S"
+    )
+
+    fh = FileHandler("clinicadl_debug.log")
```
mdiazmel:
```suggestion
    file_handler = FileHandler("clinicadl_debug.log")
```

clinicadl/utils/logwriter.py
```diff
@@ -19,37 +19,57 @@ def filter(self, record):
...
+    fh.setLevel(DEBUG)
+    fh.setFormatter(debug_formatter)
+
+    ch_info = StreamHandler(stdout)
+    ch_info.setLevel(INFO)
+    ch_info.setFormatter(info_formatter)
+
+    ch_warning = StreamHandler(stdout)
```
mdiazmel: As discussed, try to keep a single handler for the console (`console_handler`), not two (`ch_info`, `ch_warning`).

clinicadl/utils/logwriter.py
```diff
@@ -19,37 +19,57 @@ def filter(self, record):
...
+    ch_warning.setLevel(WARNING)
+    ch_warning.setFormatter(warning_formatter)
 
-    stdout = StreamHandler(stdout)
-    stdout.addFilter(StdLevelFilter())
     stderr = StreamHandler(stderr)
```
mdiazmel: To verify where error messages are sent in runtime.

mdiazmel (changes requested):
Hi,
A couple of comments to be discussed/fixed about the usage of loggers.

clinicadl/utils/logwriter.py
ravih18: I am not sure if I understand how does the `root` logger work, will it be named `clinicadl` as well ?

clinicadl/utils/logwriter.py
ravih18: It seems to work.

ravih18:
I changed the formatter as requested, I created a custom class to reduce it to a single handler.

mdiazmel (approved):
LGTM! 
Thanks @ravih18 . I'll merge it!

clinicadl/utils/logwriter.py
mdiazmel: > I am not sure if I understand how does the `root` logger work, will it be named `clinicadl` as well ?

Mi suggestion was to change the name of the log object (here called `logger`)  to `clinicadl_logger`, to improve readability. But it's not a big deal.

Issue #355, opened 2022-10-24
Quality check


acasamitjana:
Hi all,

I'm trying to run a simple example using different ADNI subjects (for example, sbj_id=sub-ADNI002S1268, session: ses-m36). The linear registration (t1-linear) seems to work well. Nonetheless, after running the extract step, the quality check fails (qc<0.1). I also checked with the MNI152NLin2009cSym template itself and I don't get more than 0.2ish on the QC table. 

I'm probably missing something (should I skull-strip first?, normalise the intensities?). I'm using the pip installation.

The instructions I'm using are:

- clinica run t1-linear InputDataset OutputDataset
- clinicadl extract image OutputDataset t1-linear
- clinicadl quality-check t1-linear OutputDataset QC_result.tsv --no-gpu

Could you provide some help on this? 
Thanks a lot!
/Adrià

camillebrianceau:
Hi, 

Thanks for your message, this is a problem that has been observed.
We now have the solution and it will be resolved in the next release (in a few weeks) but if you need it now, you can add it in your code (#357).

Don't hesitate if you have any others questions.

acasamitjana:
Thanks a lot for addressing this, very appreciated.
/Adrià

PR #356, opened 2022-11-02
Cb quality check

camillebrianceau:
This PR aims to fix the bug on quality-check pipeline (#355 ) and add a new model (DARQ).


camillebrianceau:
see (#357)

PR #357, opened 2022-11-03
Cb qc

camillebrianceau:
This PR aims to fix the bug on quality-check pipeline (https://github.com/aramis-lab/clinicadl/issues/355 ) and add a new model (DARQ)


clinicadl/quality_check/t1_linear/cli.py
```diff
@@ -21,6 +21,18 @@
 @cli_param.option.batch_size
 @cli_param.option.n_proc
 @cli_param.option.use_gpu
+@click.option(
+    "--network",
```
ravih18: Maybe we can find a more understanding name for this option, like "method" ou "algorithm". "Network" might be fine.

docs/Preprocessing/QualityCheck.md
```diff
@@ -46,6 +46,7 @@ Default value: `0.5`.
 - `--batch_size` (int) is the size of the batch used in the DataLoader. Default value: `1`.
 - `--n_proc` (int) is the number of workers used by the DataLoader. Default value: `2`.
 - `--gpu/--no-gpu` (bool) Use GPU for computing optimization. Default behaviour is to try to use a GPU and to raise an error if it is not found.
+- `use_nifti_image`(bool) is a flag allowing the pipeline to run on the nifti images and not on the extracted tensor.
```
ravih18: `--use_nifti_image`

docs/Preprocessing/QualityCheck.md
```diff
@@ -46,6 +46,7 @@ Default value: `0.5`.
...
```
ravih18: Aren't the new options (network and use-tensor) missing ?

clinicadl/utils/caps_dataset/data.py
```diff
@@ -82,7 +83,7 @@ def __init__(
             )
 
         self.elem_per_image = self.num_elem_per_image()
-        self.size = self[0]["image"].size()
+        self.size = self[0]["image"][0].size
```
ravih18: This should be working as CapsDataset manipulates tensors

clinicadl/quality_check/t1_linear/utils.py
```diff
@@ -301,37 +804,85 @@ def pt_transform(self, image):
...
-                ).clone()
-            )
+            output_images[i] = np.flip(np.rot90(dummy[16:240, 16:240]), axis=1).copy()
 
         return torch.cat(
-            [image.float().unsqueeze_(0) for image in output_images]
+            [torch.from_numpy(i).float().unsqueeze_(0) for i in output_images]
         ).unsqueeze_(0)
 
+        # # flip, resize and crop
```
ravih18: Maybe remove this part if it is not used

ravih18 (commented):
I think we should re-think how we use the CapsDataset class here, because it should anyway output a tensor and not a nifti. Maybe, if the QC is done on nifti, we could define a new data class for it, and not reuse the CapsDataset class ?
Actually I do not understand the link between CapsDataset and QCDataset.

Maybe we could check is there is a way to import the models from a library ?

PR #364, opened 2023-01-10
Fix CI pipeline in GPU machines

mdiazmel:
Add path to locate pipx in GPU machine

### DIFF
--- a/.jenkins/Jenkinsfile
+++ b/.jenkins/Jenkinsfile
@@ -22,7 +22,7 @@ pipeline {
               CONDA_ENV = "$WORKSPACE/env"
               PATH = "$HOME/.local/bin:$PATH"
               TMP_DIR = "$HOME/tmp"
-              INPUT_DATA_DIR = "/mnt/data/clinicadl_data_ci/data_ci"
+              INPUT_DATA_DIR = '/mnt/data/clinicadl_data_ci/data_ci'
             }
             stages {
               stage('Build Env') {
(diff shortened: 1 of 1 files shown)


.jenkins/Jenkinsfile
```diff
@@ -259,8 +259,9 @@ pipeline {
             environment {
               CONDA_HOME = "$HOME/miniconda3"
               CONDA_ENV = "$WORKSPACE/env"
+              PATH = "$HOME/.local/bin:$PATH"
               TMP_DIR = "$HOME/tmp"
-              INPUT_DATA_DIR = "/mnt/data/clinicadl_data_ci/data_ci"
+              INPUT_DATA_DIR = '/mnt/data/clinicadl_data_ci/data_ci'
```
NicolasGensollen, line 264: Not super important, but I don't see why these were changed from " to '.

NicolasGensollen (approved):
LGTM @mdiazmel !

.jenkins/Jenkinsfile
mdiazmel: Thanks for noticing this @NicolasGensollen, I hadn't realized that!
After investigating, it seems that the groovy linter in my editor (vscode) doesn't like the double " for this line (and it changes automatically). This [link](https://stackoverflow.com/questions/6761498/whats-the-difference-of-strings-within-single-or-double-quotes-in-groovy) gives an explanation. But I have to  confess that it is a bit disturbing to mixing double " and single ' .

NicolasGensollen: Thanks for the link @mdiazmel ! 👍

PR #373, opened 2023-02-10
Cb qc darq

camillebrianceau:
This PR aims to fix the bug on quality-check pipeline #355 and add a new model (DARQ)


clinicadl/quality_check/t1_linear/models.py
```diff
@@ -0,0 +1,636 @@
+from typing import Any, Callable, List, Optional, Type, Union
```
ravih18, line 3: I think it is good to put in the file header where does the code comes from (like it has be done in the `pytorch_ssim.py` file.

clinicadl/quality_check/t1_linear/quality_check.py
```diff
@@ -26,6 +30,9 @@ def quality_check(
     batch_size: int = 1,
     n_proc: int = 0,
     gpu: bool = True,
+    network: str = "darq",
+    use_tensor: bool = False,
+    use_uncropped_image: bool = True,
```
mdiazmel, line 35: Can you add a docstring to this function please ? Seems important to well explain the possible options for the `network` argument!

mdiazmel (approved):
LGTM! 
A small change (add docstring) and then ready to merge for me!

PR #381, opened 2023-02-15
MNT: Remove publish commands from Makefile

ghisvail:
Replaced by the tag -> draft release -> publish workflow on GitHub

I prefer removing them to not leave the impression that publishing could happen this way. It should not now that we have got working automation in place.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,10 +17,6 @@ build:
 clean.doc:
 	@$(RM) -rf site
 
-.PHONY: config.testpypi
-config.testpypi:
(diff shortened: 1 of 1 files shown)


mdiazmel:
Thanks @ghisvail . It should be also a good idea to remove all the references to the command `make publish`.
I'll start (in a new PR) by removing the Publish stage from the Jenkins pipeline.

ghisvail:
@mdiazmel Good idea. I'll let you merge these 2 PRs sequentially 👍

mdiazmel (approved):

PR #383, opened 2023-02-15
Maps manager check args before creating maps folder

camillebrianceau:
This PR allows to check if the command line arguments are correct before creating the MAPS folder.


mdiazmel (approved):
Looks good! Thanks @camillebrianceau !

PR #391, opened 2023-02-27
improve tsvtools logger

camillebrianceau:
Loggers have been improved in PR #342 but we were refactoring the tsvtools so this PR improves tsvtools logger.


mdiazmel (approved):
Looks good!

PR #392, opened 2023-03-02
add option --output_dir in get-labels

camillebrianceau:
I added an option in `clinicadl tsvtools get-labels` in order to chose another directory to save the output if the default one is not convenient.


mdiazmel (approved):
Thanks @camillebrianceau !
I merge your changes!

PR #395, opened 2023-03-08
Transition from os to pathlib

### LINKED ISSUE #262: Transition from `os` to `pathlib`
Similarly to Clinica, we could now use `pathlib` functionalities instead of `os`.

camillebrianceau:
to close issue #262


clinicadl/utils/maps_manager/iotools.py
```diff
@@ -344,15 +344,17 @@ def memReport():
...
+    import multiprocessing
     import sys
 
     import psutil
 
     print(sys.version)
     print(psutil.cpu_percent())
     print(psutil.virtual_memory())  # physical memory usage
-    pid = os.getpid()
+    process = multiprocessing.current_process()
```
ravih18, line 355: Does this do that same ? If we are not sure we can still keep os sometimes for this kind of function.

ravih18 (approved):
I am not super familiar with pathlib but LGTM.

clinicadl/utils/maps_manager/iotools.py
camillebrianceau: Yes it does, but it imports `multiprocessing` instead of `os` so I don't know if it is really relevant ...

clinicadl/generate/generate.py
```diff
@@ -202,14 +202,14 @@ def generate_trivial_dataset(
     data_df = extract_baseline(data_df)
 
     home = str(Path.home())
-    cache_clinicadl = join(home, ".cache", "clinicadl", "ressources", "masks")
+    cache_clinicadl = Path(home) / ".cache" / "clinicadl" / "ressources" / "masks"
```
mdiazmel: Here, It seems redundant with the code in the above line. I would remove the precedent line and add"
```suggestion
    home = Path.home()
    cache_clinicadl = home / ".cache" / "clinicadl" / "ressources" / "masks"
```

clinicadl/generate/generate.py
```diff
@@ -202,14 +202,14 @@ def generate_trivial_dataset(
...
     url_aramis = "https://aramislab.paris.inria.fr/files/data/masks/"
     FILE1 = RemoteFileStructure(
         filename="AAL2.tar.gz",
         url=url_aramis,
         checksum="89427970921674792481bffd2de095c8fbf49509d615e7e09e4bc6f0e0564471",
     )
-    makedirs(cache_clinicadl, exist_ok=True)
+    Path(cache_clinicadl).mkdir(parents=True, exist_ok=True)
```
mdiazmel: It's necessary to cast with `Path(...)`? It seems that `cache_clinicadl` is already a `Path` object.

clinicadl/generate/generate.py
```diff
@@ -218,7 +218,7 @@ def generate_trivial_dataset(
         )
 
     if mask_path is None:
-        if not exists(join(cache_clinicadl, "AAL2")):
+        if not (Path(cache_clinicadl) / "AAL2").is_dir():
```
mdiazmel: Same than above

clinicadl/generate/generate.py
```diff
@@ -227,7 +227,7 @@ def generate_trivial_dataset(
                 try:
                     tar_file.extractall(cache_clinicadl)
                     tar_file.close()
-                    mask_path = join(cache_clinicadl, "AAL2")
+                    mask_path = Path(cache_clinicadl) / "AAL2"
```
mdiazmel: Same

clinicadl/generate/generate.py
```diff
@@ -238,10 +238,10 @@ def generate_trivial_dataset(
                     and provide a valid path."""
                 )
         else:
-            mask_path = join(cache_clinicadl, "AAL2")
+            mask_path = Path(cache_clinicadl) / "AAL2"
```
mdiazmel: Same

clinicadl/prepare_data/prepare_data.py
```diff
@@ -54,16 +53,16 @@ def DeepLearningPrepareData(caps_directory, tsv_file, n_proc, parameters):
...
+                Path(caps_directory)
+                / container
+                / "deeplearning_prepare_data"
+                / subfolder
+                / mod_subfolder
             )
-            if not path.exists(output_file_dir):
-                os.makedirs(output_file_dir)
-            output_file = path.join(output_file_dir, filename)
+            if not Path(output_file_dir).is_dir():
```
mdiazmel: Not necessary to cast to Path
```suggestion
            if not output_file_dir.is_dir():
```

clinicadl/quality_check/t1_linear/quality_check.py
```diff
@@ -69,10 +67,10 @@ def quality_check(
     # Fetch QC model
     home = str(Path.home())
 
-    cache_clinicadl = join(home, ".cache", "clinicadl", "models")
+    cache_clinicadl = Path(home) / ".cache" / "clinicadl" / "models"
```
mdiazmel: Same comment than before, not necessary to cast `Path(home)` if the precedent line is edited.

clinicadl/quality_check/t1_linear/quality_check.py
mdiazmel, line 1: Globally, the comment is the same. You should review where the cast to a `Path` object is pertinent or not. For example, for the variable  `cache_clinicadl` isn't necessary because it's already a `Path` object.

tests/test_train_ae.py
```diff
@@ -94,14 +94,14 @@ def test_train_ae(cmdopt, tmp_path, test_name):
     flag_error = not os.system("clinicadl " + " ".join(test_input))
     assert flag_error
 
-    with open(tmp_out_dir / "maps.json", "r") as out:
-        json_data_out = json.load(out)
-    with open(ref_dir / ("maps_" + test_name) / "maps.json", "r") as ref:
-        json_data_ref = json.load(ref)
+    # with open(tmp_out_dir / "maps.json", "r") as out:
```
mdiazmel: Why theses assertions are commented?

mdiazmel (changes requested):
Hi @camillebrianceau,

Thanks for taking in charge this task, that is not trivial.

It's a good starting point. In a next iteration, I will propose to review the arguments for the functions that do I/O operations, in order to change the type. For example, variables as `output_dir`, `img_path`, `tsv_dir`, etc. should be declared in the function definition as of type `Path`  and not `str`. It's a lot of work, and as I said before, it can be done in a next PR.

Concerning this PR, globally, I think that you have to review where the cast to the `Path` object is really necessary and where isn't (I pointed our some examples in the comments of the review).

camillebrianceau:
Hi @mdiazmel,

Thank you for your review.

I agree that we need another PR to change the type of variables declared as `str` for `Path`. We also need to change the `click.option` type in some cases.

I've done a preliminary review to identify where the cast to the Path object is really necessary. If it's okay, I'd like to merge this PR and address any remaining changes in the next PR.

PR #397, opened 2023-03-13
Cb generate hypometabolic

camillebrianceau:
- This PR aims to add a pipeline that generates hypometabolic data. (for more information, see https://hal.science/hal-03835015)
- For the output filename, we will add a flag `desc-hypo-pathology-degree`. For example, the filename:
`sub-ADNI098S6658_ses-M00_trc-18FFDG_rec-uniform_pet_space-MNI152NLin2009cSym_desc-Crop_res-1x1x1_suvr-cerebellumPons2_pet.nii` would become:
`sub-ADNI098S6658_ses-M00_trc-18FFDG_rec-uniform_pet_space-MNI152NLin2009cSym_desc-Crop_res-1x1x1_suvr-cerebellumPons2_pat-ad_deg-30_pet.nii`.
- The TSV files produced are `subject_session_list.tsv`, which is the same as in the CAPS, and `data.tsv`, which contains information on the pathology and the degree of the pathology applied.


clinicadl/generate/generate.py
```diff
@@ -402,3 +406,184 @@ def generate_shepplogan_dataset(
...
+    pathology_percent: float = 60,
+    multi_cohort: bool = False,
+    uncropped_image: bool = False,
+    acq_label: str = "18FFDG",
+    suvr_reference_region: str = "cerebellumPons2",
+):
+    """
+    Generates a fully separable dataset.
+
+    Generates a dataset, based on the images of the CAPS directory, where a
```
ravih18: Description to change

clinicadl/generate/generate.py
```diff
@@ -402,3 +406,184 @@ def generate_shepplogan_dataset(
...
+    processed and image with half-left processed)
+
+    Args:
+        caps_directory: path to the CAPS directory.
+        output_dir: folder containing the synthetic dataset in CAPS format.
+        n_subjects: number of subjects in each class of the synthetic dataset.
+        tsv_path: path to tsv file of list of subjects/sessions.
+        preprocessing: preprocessing performed. Must be in ['linear', 'extensive'].
+        mask_path: path to the extracted masks to generate the two labels.
+        atrophy_percent: percentage of atrophy applied.
```
ravih18: It is not an atrophy, it is an hypometabolism

clinicadl/generate/generate.py
```diff
@@ -402,3 +406,184 @@ def generate_shepplogan_dataset(
...
+
+
+def generate_hypometabolic_dataset(
+    caps_directory: str,
+    output_dir: str,
+    n_subjects: int,
+    tsv_path: Optional[str] = None,
+    preprocessing: str = "pet-linear",
+    pathology: str = "alzheimer",
+    pathology_percent: float = 60,
```
ravih18: I think default should be 30

clinicadl/generate/generate.py
ravih18: I think more globally that the args list have not been updated

clinicadl/generate/generate.py
```diff
@@ -402,3 +406,184 @@ def generate_shepplogan_dataset(
...
+            "output_dir": output_dir,
+            "caps_dir": caps_directory,
+            "preprocessing": preprocessing,
+            "n_subjects": n_subjects,
+            "pathology": pathology,
+            "pathology_percent": pathology_percent,
+        }
+    )
+
+    dementias = ["ad", "bvftd", "lvppa", "nfvppa", "pca", "svppa"]
```
ravih18: It looks like this is not coherent with the default argument of the function

clinicadl/generate/generate.py
```diff
@@ -402,3 +406,184 @@ def generate_shepplogan_dataset(
...
     logger.info(f"Shepplogan dataset was generated at {output_dir}")
+
+
+def generate_hypometabolic_dataset(
+    caps_directory: str,
+    output_dir: str,
+    n_subjects: int,
+    tsv_path: Optional[str] = None,
+    preprocessing: str = "pet-linear",
+    pathology: str = "alzheimer",
```
ravih18: "alzheimer" or "ad" ?

clinicadl/generate/generate.py
ravih18: I think this should not be hard coded here, maybe more in the choice available in the command line.

clinicadl/generate/generate.py
```diff
@@ -402,3 +406,184 @@ def generate_shepplogan_dataset(
...
+        else:
+            mask_path = cache_clinicadl / f"mask_hypo_{pathology}.nii"
+            mask_nii = nib.load(mask_path)
+
+    # Create subjects dir
+
+    (Path(output_dir) / "subjects").mkdir(parents=True, exist_ok=True)
+
+    # Set sigma value depending on percentage
+    if pathology_percent >= 70:
```
ravih18: Maybe sigma should be an argument of the command line ? Default would be 5 then

clinicadl/generate/generate.py
```diff
@@ -402,3 +406,184 @@ def generate_shepplogan_dataset(
...
+        sigma = 9
+    elif pathology_percent >= 50:
+        sigma = 6
+    else:
+        sigma = 5
+
+    # Output tsv file
+    columns = ["participant_id", "session_id", "diagnosis", "age_bl", "sex"]
+    output_df = pd.DataFrame(columns=columns)
+    diagnosis_list = ["AD", "CN"]
```
ravih18: Is this relevant ? First all the simulated images are diseased, and then they are more demantias than just AD

clinicadl/generate/generate.py
```diff
@@ -402,3 +406,184 @@ def generate_shepplogan_dataset(
...
+        label = i
+
+        participant_id = data_df.loc[data_idx, "participant_id"]
+        session_id = data_df.loc[data_idx, "session_id"]
+        cohort = data_df.loc[data_idx, "cohort"]
+        print(participant_id)
+        print(session_id)
+        print(caps_dict[cohort])
+        print(file_type)
+        image_paths = clinica_file_reader(
```
ravih18: Can this be called only once instead of at each iteration ?

clinicadl/generate/generate.py
```diff
@@ -402,3 +406,184 @@ def generate_shepplogan_dataset(
...
+            [participant_id], [session_id], caps_dict[cohort], file_type
+        )[0]
+        image_nii = nib.load(image_paths[0])
+        image = image_nii.get_fdata()
+
+        mask_nii = resample_to_img(mask_nii, image_nii, interpolation="nearest")
```
ravih18: This can be done only once

clinicadl/generate/generate.py
```diff
@@ -402,3 +406,184 @@ def generate_shepplogan_dataset(
...
+
+        hypo_image_nii = nib.Nifti1Image(hypo_image, affine=image_nii.affine)
+        hypo_image_nii.to_filename(hypo_image_nii_dir / hypo_image_nii_filename)
+
+        # Append row to output tsv
+        row = [
+            f"sub-HYP0{i}_{pathology}-{pathology_percent}",
+            session_id,
+            diagnosis_list[label],
+            60,
```
ravih18: What is 60 ?

clinicadl/generate/generate.py
```diff
@@ -402,3 +406,184 @@ def generate_shepplogan_dataset(
...
+            "F",
```
ravih18: What is F ?

clinicadl/generate/generate.py
```diff
@@ -402,3 +406,184 @@ def generate_shepplogan_dataset(
...
+        hypo_image = hypo_synthesis(image, mask, pathology_percent, sigma)
+
+        hypo_image_nii = nib.Nifti1Image(hypo_image, affine=image_nii.affine)
+        hypo_image_nii.to_filename(hypo_image_nii_dir / hypo_image_nii_filename)
+
+        # Append row to output tsv
+        row = [
+            f"sub-HYP0{i}_{pathology}-{pathology_percent}",
+            session_id,
+            diagnosis_list[label],
```
ravih18: Do we need diagnosis in simulated dementia ? This might not be relevant

clinicadl/generate/generate.py
```diff
@@ -402,3 +406,184 @@ def generate_shepplogan_dataset(
...
+        hypo_image_nii_dir.mkdir(parents=True, exist_ok=True)
+
+        # Create atrophied image
+        hypo_image = hypo_synthesis(image, mask, pathology_percent, sigma)
+
+        hypo_image_nii = nib.Nifti1Image(hypo_image, affine=image_nii.affine)
+        hypo_image_nii.to_filename(hypo_image_nii_dir / hypo_image_nii_filename)
+
+        # Append row to output tsv
+        row = [
```
ravih18, line 594: A column with pathology and percentage could be better

clinicadl/generate/generate.py
```diff
@@ -402,3 +406,184 @@ def generate_shepplogan_dataset(
...
+            "F",
+        ]
+        row_df = pd.DataFrame([row], columns=columns)
+        output_df = pd.concat([output_df, row_df])
+
+    output_df.to_csv(join(output_dir, "data.tsv"), sep="\t", index=False)
+
+    write_missing_mods(output_dir, output_df)
+
+    logger.info(f"hypometabolic dataset was generated at {output_dir}")
```
ravih18: Make a sentence, h->H and add a dot.

ravih18: Maybe also add the parameters (pathology and percentage in this log).

clinicadl/generate/generate_hypometabolic_cli.py
```diff
@@ -0,0 +1,62 @@
...
+    "-p",
+    type=str,
+    default="ad",
+    help="Path to the extracted masks to generate the two labels. "
+    "Default will try to download masks and store them at '~/.cache/clinicadl'.",
+)
+@click.option(
+    "--pathology_percent",
+    type=float,
+    default=60.0,
```
ravih18: default should be 30

clinicadl/generate/generate_hypometabolic_cli.py
```diff
@@ -0,0 +1,62 @@
...
+    help="Percentage of pathology applied.",
```
ravih18: "Percentage of hypometabolism simulated."

clinicadl/generate/generate_hypometabolic_cli.py
```diff
@@ -0,0 +1,62 @@
...
+from clinicadl.utils import cli_param
+
+
+@click.command(name="hypometabolic", no_args_is_help=True)
+@cli_param.argument.caps_directory
+@cli_param.argument.generated_caps
+@cli_param.option.participant_list
+@cli_param.option.n_subjects
+@click.option(
+    "--pathology",
```
ravih18, line 13: maybe dementia would be more accurate.

clinicadl/generate/generate_hypometabolic_cli.py
```diff
@@ -0,0 +1,62 @@
+import click
+
+from clinicadl.utils import cli_param
+
+
+@click.command(name="hypometabolic", no_args_is_help=True)
+@cli_param.argument.caps_directory
+@cli_param.argument.generated_caps
+@cli_param.option.participant_list
+@cli_param.option.n_subjects
```
ravih18, line 10: I think this information is in participant list no ?

clinicadl/generate/generate_hypometabolic_cli.py
```diff
@@ -0,0 +1,62 @@
...
+    generate_hypometabolic_dataset(
+        caps_directory=caps_directory,
+        tsv_path=participants_tsv,
+        preprocessing="pet-linear",
+        output_dir=generated_caps_directory,
+        n_subjects=n_subjects,
+        pathology=pathology,
+        pathology_percent=pathology_percent,
+        uncropped_image=use_uncropped_image,
+        acq_label="18FFDG",
```
ravih18: These argument should be the one passed in the command line ? Otherwise we can just remove them from the CLI

clinicadl/generate/generate_utils.py
```diff
@@ -319,3 +319,12 @@ def generate_shepplogan_phantom(
     img.clip(0, 1)
 
     return img
+
+
+def hypo_synthesis(image, mask, percentage, sigma):
```
ravih18: The mask processing should be done only once when the mask is loaded, not everytime we compute a subject.

clinicadl/generate/hypometabolic.py
ravih18, line 1: This file should be removed from the repo.

docs/Preprocessing/Generate.md
ravih18, line 1: To complete.

ravih18 (changes requested):
Several changes are required:
* Some steps should be done at initialisation and not in the loop (mask generation)
* We should think about the tsv produced, the one used in other generate pipeline would not work in this case.
* The name of the output file should be debated.

clinicadl/generate/generate_hypometabolic_cli.py
camillebrianceau: Yes but `participant_list` is an option so you can chose which one you want to add (or don't add any)

clinicadl/generate/generate_hypometabolic_cli.py
```diff
@@ -0,0 +1,68 @@
...
+@cli_param.argument.generated_caps
+@cli_param.option.participant_list
+@cli_param.option.n_subjects
+@cli_param.option.n_proc
+@click.option(
+    "--pathology",
+    "-p",
+    type=click.Choice(["ad", "bvftd", "lvppa", "nfvppa", "pca", "svppa"]),
+    default="ad",
+    help="Path to the extracted masks to generate the two labels. "
```
ravih18: The help string should be updated.

clinicadl/generate/generate_hypometabolic_cli.py
```diff
@@ -0,0 +1,68 @@
...
+    "Default will try to download masks and store them at '~/.cache/clinicadl'.",
+)
+@click.option(
+    "--anomaly_degree",
+    "-ad",
```
ravih18: I think this might be confusing with the name of the pathology, maybe -anod ?

clinicadl/generate/generate_hypometabolic_cli.py
```diff
@@ -0,0 +1,68 @@
...
+    type=float,
+    default=30.0,
+    help="Degrees of hypo-metabolism applied (in percent)",
+)
+@click.option(
+    "--sigma",
+    type=int,
+    default=5,
+    help="sigma ?",
```
ravih18: it is the parameter of the gaussian filter used for smoothing.

clinicadl/generate/generate.py
```diff
@@ -432,3 +438,180 @@ def generate_shepplogan_dataset(
...
+    Returns
+    -------
+    Folder structure where images are stored in CAPS format.
+
+
+    Raises
+    ------
+    IndexError: if `n_subjects` is higher than the length of the TSV file at `tsv_path`.
+    """
+    from pathlib import Path
```
ravih18: already imported

clinicadl/generate/generate.py
```diff
@@ -432,3 +438,180 @@ def generate_shepplogan_dataset(
...
+
+    from clinicadl.utils.exceptions import DownloadError
```
ravih18: should be imported l22 ?

clinicadl/generate/generate.py
```diff
@@ -432,3 +438,180 @@ def generate_shepplogan_dataset(
...
+
+    if n_subjects > len(data_df):
+        raise IndexError(
+            f"The number of subjects {n_subjects} cannot be higher "
+            f"than the number of subjects in the baseline dataset of size {len(data_df)}"
+            f"Please add the '--n_subjects' option and re-run the command."
+        )
+
+    mask_path = None
+    if mask_path is None:
```
ravih18: mask_path will always be None, does it needs to be tested ?

clinicadl/generate/generate.py
```diff
@@ -432,3 +438,180 @@ def generate_shepplogan_dataset(
...
+            checksum="drgjeryt",
+        )
+        cache_clinicadl.mkdir(parents=True, exist_ok=True)
+
+        if not (cache_clinicadl / f"mask_hypo_{pathology}.nii").is_file():
+            logger.info(f"Downloading {pathology} masks...")
+            # mask_path = fetch_file(FILE1, cache_clinicadl)
+            try:
+                mask_path = fetch_file(FILE1, cache_clinicadl)
+                mask_nii = nib.load(mask_path)
```
ravih18: this can be outside of the if..else

clinicadl/generate/generate.py
```diff
@@ -432,3 +438,180 @@ def generate_shepplogan_dataset(
...
+            except:
+                DownloadError(
+                    """Unable to download masks, please download them
+                    manually at https://aramislab.paris.inria.fr/files/data/masks/
+                    and provide a valid path."""
+                )
+        else:
+            mask_path = cache_clinicadl / f"mask_hypo_{pathology}.nii"
+            mask_nii = nib.load(mask_path)
```
ravih18: this can be outside of the if..else

clinicadl/generate/generate.py
```diff
@@ -432,3 +438,180 @@ def generate_shepplogan_dataset(
...
+    output_df = pd.DataFrame(columns=columns)
+    participants = [data_df.loc[i, "participant_id"] for i in range(n_subjects)]
+    sessions = [data_df.loc[i, "session_id"] for i in range(n_subjects)]
+    cohort = caps_directory
+
+    images_paths = clinica_file_reader(participants, sessions, cohort, file_type)[0]
+
+    image_nii = nib.load(images_paths[0])
+
+    mask_resample_nii = resample_to_img(mask_nii, image_nii, interpolation="nearest")
```
ravih18, line 563: maybe these two steps can also be in the mask_processing function

clinicadl/generate/generate.py
```diff
@@ -432,3 +438,180 @@ def generate_shepplogan_dataset(
...
+    mask = mask_resample_nii.get_fdata()
+    mask = mask_processing(mask, anomaly_degree, sigma)
+
+    # Create subjects dir
+    (output_dir / "subjects").mkdir(parents=True, exist_ok=True)
+
+    def generate_hypometabolic_image(i, output_df):
+        # for i in range(n_subjects):
```
ravih18: remove this comment

clinicadl/generate/generate.py
```diff
@@ -432,3 +438,180 @@ def generate_shepplogan_dataset(
...
+        image = image_nii.get_fdata()
+
+        input_filename = image_path.name
+        filename_pattern = "_".join(input_filename.split("_")[2::])
+
+        hypo_image_nii_dir = (
+            output_dir / "subjects" / f"sub-HYPO{i}" / sessions[i] / preprocessing
+        )
+        hypo_image_nii_filename = (
+            f"sub-HYPO{i}_{pathology}-{anomaly_degree}_{sessions[i]}_{filename_pattern}"
```
ravih18: I think we should still have the original dataset ID since the goal is to have a patient specific abnormal image.

ravih18 (changes requested):
We still need to agree on the output file name. In my opinion it should follow the original dataset name, with a flag `desc-hypo_ad_30` for instance for 30% of AD.
Indeed the goal is to be able to find to which subject does the abnormal image belongs.

clinicadl/generate/generate.py
camillebrianceau: It would require to add the `image_nii` as a parameters of the function ..

PR #400, opened 2023-03-20
PyTorch Profiler

ncassereau:
Hi,

This new PR would add some code to allow the usage of the Pytorch profiler. This change would not help most user, but would allow ClinicaDL's dev team to evaluate changes. Some users might even use it to tune --n_proc or understand better the computational aspect of their own model.

The pytorch profiler can be buggy with some recent version of Pytorch (my tests were done with Pytorch 1.11.0, I do not know about Pytorch 2.0, but I do know that it's buggy with Pytorch 1.13). It gives you the usage of your CPU, of your GPU, the ratio of communications, GPU computations, data loading, CPU-GPU transfers, etc.

We could add other flags and parameters to tune the scheduling of the profiler, or the output filename, but I'm not sure it's worth the trouble, what do you think ?

Considering ClinicaDL is very limited by memory, enabling the profile_memory flag is probably very worth it. The 3 other flags (record_shapes, with_stacks, with_flops) are probably not necessary and might slow down the code even further without any obvious added value. Record_shapes and with_stacks might help with debugging. With_flops might help with the hardware evaluation.

With the future plan of adding Distributed Data Parallelism, I've added a commented-out line which would make sure all processes have the same filename. Because I am using the time of the execution in order not to overwrite files, there is a tiny probability that two processes have different filenames, so I'm adding a broadcast step. Maybe using the datetime is not a good idea. I can either change this naming convention, or will definitely add this line in the DDP pull request.

For people who wants to know what it looks like, I've put on Jean Zay some traces in $krk_ALL_CCFRWORK/profiler. They can be opened with Tensorboard.

### DIFF
--- a/clinicadl/resources/config/train_config.toml
+++ b/clinicadl/resources/config/train_config.toml
@@ -38,7 +38,6 @@ gpu = true
 n_proc = 2
 batch_size = 8
 evaluation_steps = 0
-profiler = false
 
 [Reproducibility]
 seed = 0
@@ -73,3 +72,4 @@ weight_decay = 1e-4
 patience = 0
 tolerance = 0.0
 accumulation_steps = 1
+profiler = false
--- a/clinicadl/resources/config/train_config.toml
+++ b/clinicadl/resources/config/train_config.toml
@@ -38,6 +38,7 @@ gpu = true
 n_proc = 2
 batch_size = 8
 evaluation_steps = 0
+profiler = False
 
 [Reproducibility]
 seed = 0
--- a/clinicadl/resources/config/train_config.toml
+++ b/clinicadl/resources/config/train_config.toml
@@ -38,7 +38,7 @@ gpu = true
 n_proc = 2
 batch_size = 8
 evaluation_steps = 0
-profiler = False
+profiler = false
 
 [Reproducibility]
 seed = 0
--- a/clinicadl/train/train_option.py
+++ b/clinicadl/train/train_option.py
@@ -261,7 +261,8 @@ profiler = cli_param.option_group.optimization_group.option(
     "--profiler/--no-profiler",
     type=bool,
     default=None,
-    help="Use `--profiler` to enable Pytorch profiler for the first few steps.",
+    help="Use `--profiler` to enable Pytorch profiler for the first 30 steps after a short warmup. "
+    "It will make an execution trace and some statistics about the CPU and GPU usage.",
 )
 # transfer learning
 transfer_path = cli_param.option_group.transfer_learning_group.option(
--- a/clinicadl/train/train_option.py
+++ b/clinicadl/train/train_option.py
@@ -48,7 +48,7 @@ profiler = cli_param.option_group.computational_group.option(
     "--profiler/--no-profiler",
     type=bool,
     default=None,
-    help="Use `--profiler` to enable Pytorch profiler for the first few steps."
+    help="Use `--profiler` to enable Pytorch profiler for the first few steps.",
 )
 # Reproducibility
 seed = cli_param.option_group.reproducibility_group.option(
--- a/clinicadl/train/train_option.py
+++ b/clinicadl/train/train_option.py
@@ -44,6 +44,12 @@ evaluation_steps = cli_param.option_group.computational_group.option(
     help="Fix the number of iterations to perform before computing an evaluation. Default will only "
     "perform one evaluation at the end of each epoch.",
 )
+profiler = cli_param.option_group.computational_group.option(
+    "--profiler/--no-profiler",
+    type=bool,
+    default=None,
+    help="Use `--profiler` to enable Pytorch profiler for the first few steps."
+)
 # Reproducibility
 seed = cli_param.option_group.reproducibility_group.option(
     "--seed",
--- a/clinicadl/train/train_option.py
+++ b/clinicadl/train/train_option.py
@@ -44,12 +44,6 @@ evaluation_steps = cli_param.option_group.computational_group.option(
     help="Fix the number of iterations to perform before computing an evaluation. Default will only "
     "perform one evaluation at the end of each epoch.",
 )
-profiler = cli_param.option_group.computational_group.option(
-    "--profiler/--no-profiler",
-    type=bool,
-    default=None,
-    help="Use `--profiler` to enable Pytorch profiler for the first few steps.",
-)
 # Reproducibility
 seed = cli_param.option_group.reproducibility_group.option(
     "--seed",
@@ -262,6 +256,12 @@ accumulation_steps = cli_param.option_group.optimization_group.option(
     # default=1,
     help="Accumulates gradients during the given number of iterations before performing the weight update "
     "in order to virtually increase the size of the batch.",
+)
+profiler = cli_param.option_group.optimization_group.option(
+    "--profiler/--no-profiler",
+    type=bool,
+    default=None,
+    help="Use `--profiler` to enable Pytorch profiler for the first few steps.",
 )
 # transfer learning
 transfer_path = cli_param.option_group.transfer_learning_group.option(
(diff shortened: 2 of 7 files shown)


mdiazmel:
Thanks @ncassereau-idris for this awesome feature!

> The pytorch profiler can be buggy with some recent version of Pytorch (my tests were done with Pytorch 1.11.0, I do not know about Pytorch 2.0, but I do know that it's buggy with Pytorch 1.13). It gives you the usage of your CPU, of your GPU, the ratio of communications, GPU computations, data loading, CPU-GPU transfers, etc.

Concerning the support of Pytorch 2.0 in ClinicaDL, it's not in our roadmap, at least in the near future. Currently we support `Pytorch = ^1.8.0`.  We hope that it works well at least with the current stable versions (1.xx).

clinicadl/train/train_option.py
```diff
@@ -44,6 +44,12 @@
     help="Fix the number of iterations to perform before computing an evaluation. Default will only "
     "perform one evaluation at the end of each epoch.",
 )
+profiler = cli_param.option_group.computational_group.option(
+    "--profiler/--no-profiler",
+    type=bool,
+    default=None,
+    help="Use `--profiler` to enable Pytorch profiler for the first few steps.",
```
mdiazmel: What does mean exactly "enable Pytorch profiles for the first few steps"?

camillebrianceau (approved):
LGTM ! Thanks Nathan !

About adding other flags and parameters, I agree I think it doesn't worth the trouble.

mdiazmel (approved):
Thx for this contribution, @ncassereau-idris 

The tests doesn't pass because we need to modify our reference data. In fact, all the parameters of the training are saved in a json file (`maps.json`) and we compare the contents of this file during the tests. As a new parameter is introduced with this PR (`--profiler`) we need to update our reference data to include it (this will be done once the PR is merged).

About the flags to tune the scheduling of the profiler, I think that you have more expertise on the topic to decide which one to include. Maybe, the  output filename could be also interesting (I suppose that it's the output filename containing the stack traces, isn't?)... btw, do you recommend a particular tool to visualize this kind of file?

I'm agree that the `profile_memory` seems to be the most useful feature for ClinicaDL users, so let's keep like that.

Nothing more to add, I propose to merge once we agree if it's worth to include the output filename or not.

clinicadl/train/train_option.py
```diff
@@ -44,6 +44,12 @@
     help="Fix the number of iterations to perform before computing an evaluation. Default will only "
     "perform one evaluation at the end of each epoch.",
 )
+profiler = cli_param.option_group.computational_group.option(
+    "--profiler/--no-profiler",
+    type=bool,
+    default=None,
```
ravih18: Default should not be `False`?

clinicadl/utils/maps_manager/maps_manager.py
```diff
@@ -761,6 +761,39 @@ def _train_multi(self, split_list: List[int] = None, resume: bool = False):
 
             self._erase_tmp(split)
 
+    def _init_profiler(self):
```
ravih18: This function should be defined at the end of the maps manager, where there are other initialization methods.

clinicadl/resources/config/train_config.toml
```diff
@@ -38,6 +38,7 @@ gpu = true
 n_proc = 2
 batch_size = 8
 evaluation_steps = 0
+profiler = false
```
ravih18: I am not sure that `Computational` is the best category for this option. Maybe this can be in the `Model` category ? or `Optimization` ? Or here also might be fine.

ravih18 (commented):
Hello, thanks for this new feature ! I just have a few small comments before we merge it.

clinicadl/resources/config/train_config.toml
mdiazmel: +1 for `Optimization`!

clinicadl/train/train_option.py
camillebrianceau: I think it's `None` for the other options too

ravih18: You are right.

clinicadl/train/train_option.py
ncassereau: We can decide for how long the profiling should be enabled. It could be chosen through flags but that seems unnecessary. In this PR, I suggest 30 steps with a small warm-up without any repetition.

30 steps is enough to have multiple dataloader cycles (because I assume that people will seldom have a num_workers higher than 10 CPU per GPU), but not too many to have a file which is not too big (around 100 MB). When profiling with PyTorch,  you have a fine-grained trace of the execution on the GPU as well as the CPU, so that makes a pretty big file if you don't stop soon enough. It also slows down the execution a tiny bit because the profiler has to check on the GPU to see what it is doing to gather some information and compute some statistics.

Furthermore, even though we can keep going for an epoch or even the whole training if we really wanted, the added-value of having the trace beyond the first few cycles is slim to none, so we might as well schedule a limited amount of steps. When it is reached, we can keep the training going, because the profiler will write all its data on the disk and then shutdown and become some kind of nullcontext.

Those parameters are provided with the following line: https://github.com/aramis-lab/clinicadl/pull/400/files#diff-96d0820a215b2ed6bda3270e00c5b0725948a0ebc3f0f0d6d071e7b2540983c3R783

I could change those numbers obviously.


For the filename, it is actually the name of the folder of all the traces registered for this training (if you are using distributed training, which we will be with the DDP PR, then we will have as many traces as we have processes). The file itself is not something we really look at, it's for tensorboard to handle. It's actually the hostname, the PID and the timestamps, but it's completely invisible to the user. In Tensorboard, we only see the folder name as if it was one giant big file.

Currently the name is defined (here https://github.com/aramis-lab/clinicadl/pull/400/files#diff-96d0820a215b2ed6bda3270e00c5b0725948a0ebc3f0f0d6d071e7b2540983c3R776-R777) as the default filename (clinica_dl) and the time (HH:MM:SS) in order not to overwrite files. All of that will be stored in a "profiler" folder so that everything is nice and clean. This profiler folder will be stored in the folder where the clinicadl command is called, I could put that in the output directory with the maps.json, no idea whether that's relevant or not. If we want to tune it, I could add something like:
```python
profiler_filename = cli_param.option_group.optimization_group.option(
    "--profiler-filename",
    type=str,
    default=None, # default could be "clinicadl" or whatever
    help="Defines the name where profiler traces will be stored",
)
```
and
```python
filename = [Path("profiler") / f"{self.profiler_filename}_{datetime.now().strftime("%H:%M:%S")}"]
```
but but that's an additional argument in the CLI. As you wish really :)

The traces can then analyzed with Tensorboard. It will compute the statistics and have a nice interface to check the trace, the memory profile or any global statistic like the tensor core usage, the GPU utilization, etc.

clinicadl/utils/maps_manager/maps_manager.py
ncassereau: Implemented in 813b2d98b3757e009a5787db984f5e5cfdea58e9

clinicadl/resources/config/train_config.toml
ncassereau: Implemented in 622925e6da910549ca537da49539866240e412e3

clinicadl/train/train_option.py
camillebrianceau: I think that changing the name is not worth the trouble, but the profiler folder could be located somewhere else, perhaps in the output directory. 
We could consider merging it in this way and see how it is over time ?

mdiazmel: Thanks for all this info @ncassereau-idris .

I propose to follow Camille's suggestion: to put the profiler folder in the output directory (by default) and then merge the PR. 

We'll see how it works and if it's convenient for a day-to-day usage. Otherwise, a new PR, adding a custom location for this folder, might be proposed.

ncassereau: Implemented in ffdd8a5f42c398b5eb013b91899b984bfb86f8aa

PR #401, opened 2023-03-20
Automatic Mixed Precision

ncassereau:
Automatic Mixed Precision enables Pytorch to promote some tensors to 16 bits instead of 32 bits. This saves memory and speeds up computation a tiny bit. On NVIDIA V100 GPU, it also enables the usage of tensor cores because V100 tensor cores have a limited list of data types they accept. NVIDIA A100 do not have this issue as much because they can work with fp32 tensors.

Because fp16 tensors do not have as many bits of exponent as fp32 (5 bits for fp16, 8 for fp32), I added Pytorch's GradScaler to avoid underflows and overflows. I did not include the GradScaler state_dict saving and loading. It is a thing ; not doing it will not damage the training process in any way, but technically it can change it. This only affects programs where the training is resumed. Not saving gradscaler states means that doing all your epochs in a row and a stop-and-resume approach will have different results. By different results, I mean like when you change the seed, not that it will make a whole new training process. Still, you might consider it an obstacle to reproductibility so GradScaler state_dict could be saved. That's a whole new file though. Or maybe we can save it in the same file as the optimizer. As you wish really.

With Pytorch's AMP, it is possible to choose the promoted data type. By default, it's fp16 but we could choose bf16. Bfloat16 is a 16 bits format which trades bits of mantissa for additional bits of exponents. It avoids the necessity of adding a GradScaler (because it has 8 bits of exponent, just like fp32), and some research shows that it has nice convergence properties for large models (Source: [Huggingface blogpost](https://huggingface.co/blog/bloom-megatron-deepspeed#bf16optimizer)). However, not all hardware (for instance V100 tensor cores, source [here](https://images.nvidia.com/aem-dam/en-zz/Solutions/design-visualization/technologies/turing-architecture/NVIDIA-Turing-Architecture-Whitepaper.pdf)) tolerate fp16 and it seems that it is especially useful for large language models which are orders of magnitude bigger than what is trained with ClinicaDL. One possibility is to have a flag which sets this dtype to either fp16 or bf16, but it's probably very niche, and if the training is done in bf16 on A100 GPU and you need to use it for inference on a V100 GPU, that could be an issue because some weights of the network might be too big for fp16.

Finally, I added the set_to_none=True keyword in zero_grad methods. It literally has nothing to do with mixed precision, but it might speedup things just a tiny bit. It's supposed to become the default in Pytorch 2.0, but setting it would enable this optimization on older versions of Pytorch. I could remove that though if it's inappropriate regarding the AMP pull request.


clinicadl/utils/task_manager/task_manager.py
```diff
@@ -192,6 +195,13 @@ def test(
         total_loss = {}
         with torch.no_grad():
             for i, data in enumerate(dataloader):
+                with autocast(enabled=amp):
+                    outputs, loss_dict = model.compute_outputs_and_loss(
+                        data, criterion, use_labels=use_labels
+                    )
+                total_loss += loss_dict["loss"].float().item()
```
camillebrianceau: `total_loss` is a dictionary so this line isn't working.. Is this line necessary ?

camillebrianceau (commented):
Thanks a lot :)
Can you please add some documentation for that ?

PR #402, opened 2023-03-20
Distributed Data Parallelism

ncassereau:
DDP allows the usage of multiple GPU to compute a larger batch of data. It allows us to either increase the size of the model or increase the batch size or increase our data memory footprint by for instance removing downsampling, etc.

DDP is more efficient and more flexible than DP. The usage of two GPU on the same node, or two different nodes of GPU are both covered by this PR.

To use DDP, each process of clinicadl needs to know the world size, its rank, as well as the master address and master port. For now, the cluster resolver I suggest only supports the SLURM scheduler, but if this PR is successful, we will add other cluster resolver in the future.

I also suggest the introduction of ZeRO (Zero Redundancy Optimizer). This technology from DeepSpeed (Microsoft) shards optimizer states along the data parallelism dimension. ZeRO is most optimal with the deepspeed library but that would add another dependency so that's a discussion for another day. Pytorch does have a small implementation. It only has the first stage of ZeRO (only optimizer states sharding, stage 2 introduces gradients sharding and stage 3 parameters sharding). Also, unlike DeepSpeed version, Pytorch implementation of ZeRO does increase the volume of communication required to synchronize devices. This is due to the fact that Pytorch probably wanted to limit the amount of code one would need to change to add this feature. As a result, they did not discard the second part of the gradient all-reduce (this all-gather, the first part being a ReduceScatter, or MPI_All_to_all) which is unnecessary with ZeRO. On the other hand, it indeed makes ZeRO usage painless, it's only a few additional lines of code. This feature reduces the memory footprint of the optimizer. The more GPU you have, the less memory per GPU you need. With a 300M parameters network and Automatic Mixed Precision, it reduced the amount of memory needed from 17.1GB to 15.5GB with 4 GPU. So that's neat!


clinicadl/utils/task_manager/task_manager.py
```diff
@@ -215,14 +217,27 @@ def test(
...
+        results_df.reset_index(inplace=True, drop=True)
 
         if not use_labels:
             metrics_dict = None
         else:
             metrics_dict = self.compute_metrics(results_df)
+            handles = []
             for loss_component in total_loss.keys():
-                metrics_dict[loss_component] = total_loss[loss_component]
+                handles.append(
```
ncassereau: Actually it's not very smart to do it asynchronously, I'll revert that change

clinicadl/utils/cli_param/train_option.py
```diff
@@ -44,6 +44,12 @@
     help="Fix the number of iterations to perform before computing an evaluation. Default will only "
     "perform one evaluation at the end of each epoch.",
 )
+zero = cli_param.option_group.computational_group.option(
```
ravih18: I would choose an option name that is more transparent. Moreover it might be good to specify the default value.

clinicadl/utils/task_manager/task_manager.py
```diff
@@ -133,7 +134,11 @@ def output_size(
     @staticmethod
     @abstractmethod
     def generate_sampler(
-        dataset: CapsDataset, sampler_option: str = "random", n_bins: int = 5
+        dataset: CapsDataset,
+        sampler_option: str = "random",
+        n_bins: int = 5,
+        world_size: Optional[int] = None,
```
ravih18: I am not sure this is the best name for this variable, maybe something more transparent such as `dp_degree` would be better.

clinicadl/utils/maps_manager/maps_manager.py
```diff
@@ -76,6 +80,7 @@ def __init__(
                 )
             test_parameters = self.get_parameters()
             self.parameters = change_str_to_path(test_parameters)
+            init_ddp(gpu=self.parameters["gpu"], logger=logger)
```
ravih18: This can be outside the if..else as it is executed in both conditions.

ravih18 (commented):
Nice work ! looks fine to me. only a few small comments.

However it would be great to add more docstrings to class and methods implemented in `clinicadl/utils/maps_manager/cluster/` files as it will be easier in the future to understand what they do, what they are used for to ease the maintenance.

clinicadl/utils/cli_param/train_option.py
ncassereau: Corrected in https://github.com/aramis-lab/clinicadl/pull/402/commits/d14c9c0462a7584fb7000bf5491fe17d8fb92943

clinicadl/utils/task_manager/task_manager.py
ncassereau: Corrected in https://github.com/aramis-lab/clinicadl/pull/402/commits/f74c898715f9a272f87b9a3737732ed3e3d6b406

clinicadl/utils/maps_manager/maps_manager.py
ncassereau: Corrected in https://github.com/aramis-lab/clinicadl/pull/402/commits/2c4f3b78cdfe68ab2d3de93748be510b57b1c83a

clinicadl/resources/config/train_config.toml
```diff
@@ -39,6 +39,7 @@ gpu = true
 n_proc = 2
 batch_size = 8
 evaluation_steps = 0
+fullyshardeddataparallel = false
```
ravih18: the convention we use is to sepatate words with underscore: `fully_sharded_data_parallel = false`

ncassereau: solved in cad088d

ravih18 (approved):

PR #403, opened 2023-03-21
Add new architectures

sophieloiz:
Add two new CNN architectures : SE-CNN and Resnet 3D


clinicadl/utils/network/cnn/models.py
```diff
@@ -285,3 +287,91 @@ def get_dimension():
...
+    @staticmethod
+    def get_dimension():
+        return "3D"
+
+    @staticmethod
+    def get_task():
+        return ["classification", "regression"]
+
+
+class SE_CNN(CNN):
```
ravih18: Maybe make this name more comprehensive, like `SqueezeExcitationCNN` .

PR #416, opened 2023-04-26
add caps_directory option in get-labels

camillebrianceau:
I just added a `--caps_directory` option to match with the `merge-tsv` pipeline.


ravih18 (approved):
Don't forget to update the doc accordingly

PR #417, opened 2023-04-26
Data augmentation with torchio

sophieloiz:
Add different transformation such as Random Motion, RandomGhosting, RandomBlur of torchio as data augmentation.


ravih18 (commented):
Looks good, do you think that it can be clearer to separate all the data augmentation in another file ?

camillebrianceau:
I think we can let it this way because a lot of things will be separated during the refactoring ...

ravih18 (approved):
Can be merged

PR #419, opened 2023-04-26
Add generate motion using torchio

sophieloiz:
Corrupt a CAPS by simulating motion artefact on MRIs. The image based motion simulation is made using the torchio implementation of Shaw's paper.


clinicadl/generate/generate.py
```diff
@@ -5,13 +5,15 @@
 """
 import tarfile
 from logging import getLogger
+from os.path import join
```
ravih18: do we need os.path.join ?

clinicadl/generate/generate.py
```diff
@@ -655,3 +657,115 @@ def generate_hypometabolic_image(subject_id, output_df):
...
+    columns = ["participant_id", "session_id", "diagnosis"]
+    output_df = pd.DataFrame(columns=columns)
+
+    # Find appropriate preprocessing file type
+    file_type = find_file_type(
+        preprocessing, uncropped_image, tracer, suvr_reference_region
+    )
+
+    for data_idx in range(len(data_df)):
+        participant_id = data_df.loc[data_idx, "participant_id"]
```
ravih18, line 722: Isn't there a `get_subject_session` function that we usually use for this ?

ravih18 (commented):
Few comments:
* It would be great to directly add the parallelized version of this as for other generation pipelines
* Test and doc are missing

ravih18 (commented):
Few comments:
* It would be great to directly add the parallelized version of this as for other generation pipelines
* Test and doc are missing

clinicadl/generate/generate.py
camillebrianceau: no

clinicadl/generate/generate.py
camillebrianceau: For the `generate` I think it is better like this because there is still the cohort that is not an output of `get_subject_session`

camillebrianceau:
Tests and parallelized version have been added !

clinicadl/generate/generate.py
ravih18: ok

clinicadl/generate/generate.py
```diff
@@ -655,3 +657,120 @@ def generate_hypometabolic_image(subject_id, output_df):
...
+        cohort = data_df.loc[data_idx, "cohort"]
+        image_path = Path(
+            clinica_file_reader(
+                [participant_id], [session_id], caps_dict[cohort], file_type
+            )[0][0]
+        )
+        input_filename = image_path.name
+        filename_pattern = "_".join(input_filename.split("_")[2::])
+
+        trivial_image_nii_dir = (
```
ravih18: Maybe change the name with a more explicit variable: `motion_image` seems more appropriate.

clinicadl/generate/generate.py
```diff
@@ -655,3 +657,120 @@ def generate_hypometabolic_image(subject_id, output_df):
...
+            output_dir
+            / "subjects"
+            / f"{participant_id}-RM{data_idx}"
+            / session_id
+            / preprocessing
+        )
+        trivial_image_nii_filename = (
```
ravih18: Same for here and other variables after that

ravih18 (approved):

PR #431, opened 2023-05-23
Cb fine tuning

camillebrianceau:
This PR aims to add a functionality that enables the freezing of specific layers within the model for fine-tuning. The proposed option, `--nb_unfrozen_layer`, allows the user to specify the number of fully connected layers that will undergo retraining.

At present, this is only a preliminary version of the PR, based on Sophie's work. Currently, our assumption is that the unfrozen layers are always FC layers. However, to ensure accuracy, additional code needs to be implemented to confirm that these are indeed FC layers.

There are several points that require discussion, such as:
- Determining an appropriate name for the option.
- Deciding whether to keep all FC layers unfrozen by default. If so, the option could be a boolean, such as `--freeze_conv`.


nburgos:
Thanks @camillebrianceau & @sophieloiz!
Would it be a lot of work to give the possibility to unfreeze other layers than the FC ones?
If we stick to the FC ones, calling the option `--nb_unfrozen_fc_layer` would make this fact more explicit.

sophieloiz:
In fact, the possibility to unfreeze layers other than fully connected (FC) layers has already been implemented. When using the `--nb_unfrozen_layer` parameter, you have the flexibility to choose the number of layers to unfreeze within the model. For instance, if you set it to 4, you will unfreeze the three FC layers as well as the last convolutional layer.

nburgos:
Perfect! You can forget my comment then.

PR #434, opened 2023-05-26
add __init__.py when missing

camillebrianceau:
Not sure it is necessary ?


ghisvail:
You're gonna need them to do `from clincadl.x.y import cli` at some point.

ravih18 (approved):
Nothing much to review here 😆

PR #453, opened 2023-06-27
Cb wandb

camillebrianceau:
This PR aims to add an option to track your training experiments by introducing the flag `--track_exp mlflow/wandb`. 
WandB and Mlflow are optional dependencies of ClinicaDL, as suggested by @mdiazmel. 

The documentation needs to be added, but it is ready for review! :)

I'll add the data for the CI once the first review is done.


mdiazmel (changes requested):
Hi @camillebrianceau ,

I'm definitely against adding experiment trackers to the code base of ClinicaDL. They will be added as optional dependencies (as the docs dependencies in the pyproject file) and you should find a way to import the modules not directly in the `maps manager` but in the specific functions/classes where they are really used. If we decide to integrated, W&B and MLFlow should be optional packages not necessary to the *standard* way  of using ClinicaDL.

clinicadl/utils/tracking_exp.py
```diff
@@ -0,0 +1,133 @@
...
+            experiment_id = self._mlflow.create_experiment(
+                f"clinicadl-{maps_name}",
+                artifact_location=Path.cwd().joinpath("mlruns").as_uri(),
+            )
+
+        except mlflow.exceptions.MlflowException:
+            self._mlflow.set_experiment(self.experiment_name)
+
+        self._mlflow.start_run(experiment_id=experiment_id, run_name=f"split-{split}")
+        self._mlflow.autolog()
```
mdiazmel, line 131: Have you tested if the autolog functionality of _mlflow_ works well with Pytorch? [According to the documentation](https://mlflow.org/docs/latest/python_api/mlflow.pytorch.html), it's only available for a range of versions. We should probably reflects this in our dependency file.

pyproject.toml
```diff
@@ -52,6 +52,12 @@ pytest-timeout = "*"
 pytest-xdist = "*"
 pytest-cov = "^3.0.0"
 
+[tool.poetry.group.tracking]
+optional = true
+[tool.poetry.group.tracking.dependencies]
+wandb = "^0.15.4"
+mlflow = "*"
```
mdiazmel: Use `mlflow = ^2.6.0` which currently support autologin for `1.6.0 <= torch <= 2.0.1` . Other mlflow versions support different Pytorch versions.

mdiazmel (changes requested):
Seems ok. I think that it's important to freeze the mlflow version in order to guarantee compatibility with the autolog function (which depends also the the Pytorch version). 
Let me know when the suggested changes will be implemented, I will launch a new mlflow server to test if the feature works as planned.

mdiazmel:
We still need to verify why the GPU tests fail for this PR.

clinicadl/utils/cli_param/train_option.py
```diff
@@ -282,6 +282,17 @@
...
+track_exp = cli_param.option_group.optimization_group.option(
+    "--track_exp",
+    "-te",
+    type=click.Choice(
+        [
+            "wandb",
+            "mlflow",
+        ]
+    ),
+    help="Use `--track_exp` to enable wandb to track your experience.",
```
ravih18: The help string can be improved (wand and mlflow). Maybe explain a bit what do you mean by tracking, is experience a training job ? etc

clinicadl/utils/tracking_exp.py
```diff
@@ -0,0 +1,133 @@
...
+
+def wandb_is_available():
+    return importlib.util.find_spec("wandb")
+
+
+def mlflow_is_available():
+    return importlib.util.find_spec("mlflow") is not None
+
+
+class Tracker:
```
ravih18, line 20: add doc strings if possible.

clinicadl/utils/tracking_exp.py
```diff
@@ -0,0 +1,133 @@
...
+
+        if track_exp == "wandb":
+            tracker.log(metrics_dict)
+            return metrics_dict
+        elif track_exp == "mlflow":
+            tracker.log_metrics(metrics_dict)
+            return metrics_dict
+
+
+class WandB_class(Tracker):
```
ravih18: i am not sure it is necessary to have `class` in the class name, maybe `MLflow_handler` or something like that would sound better.

clinicadl/utils/tracking_exp.py
```diff
@@ -0,0 +1,133 @@
...
+            config=config,
+            save_code=True,
+            group=maps_name,
+            mode="online",
+            name=f"split-{split}",
+            reinit=True,
+        )
+
+
+class Mlflow_class(Tracker):
```
ravih18: same comment

ravih18 (approved):

PR #457, opened 2023-07-20
Generate - add generate trivial contrast - motion and noise 

StephaneMbll:
Provide implementation for generation of synthetic dataset corrupted by contrast, motion, and noise.
Add commands `generate [trivial_contrast | trivial_noise]` 
Based on _TorchIO_ functions
Add non functional architecture for `generate trivial_artifacts`, aiming at combining contrast noise and motion generation in a single command, with possibility of artifacts combination. **(non implemented)**

PR #460, opened 2023-08-09
Generate artifacts

camillebrianceau:
I made some changes to Stephane PR (PR #457 ), but I made some mistakes, so I recreated the PR. This one is the correct version.

The goal is to add some generating options and regroup everything in a unified pipeline, where you can combine several artifacts. 
The pipeline is as follows: `clinicadl generate artifacts --noise --option_noise --contrast --option_contrast`.
We chose not to be concerned about the order of the artifacts.

We still need to define the file names to comply with the BIDS format and to add data to the CI.


clinicadl/generate/artefacts/generate_trivial_contrast_cli.py
```diff
@@ -0,0 +1,191 @@
...
+    find_file_type,
+    load_and_check_tsv,
+    write_missing_mods,
+)
+from clinicadl.utils import cli_param
+from clinicadl.utils.caps_dataset.data import CapsDataset
+from clinicadl.utils.maps_manager.iotools import commandline_to_json
+
+
+def generate_contrast_dataset(
```
ravih18: It is better to define these function in a separate file and keep the `..._cli.py` file only to execute the command line instructions.

clinicadl/generate/artefacts/generate_trivial_motion_cli.py
```diff
@@ -0,0 +1,217 @@
...
+    write_missing_mods,
+)
+from clinicadl.utils import cli_param
+from clinicadl.utils.caps_dataset.data import CapsDataset
+from clinicadl.utils.maps_manager.iotools import commandline_to_json
+
+logger = getLogger("clinicadl.generate")
+
+
+def generate_motion_dataset(
```
ravih18: It is better to define these function in a separate file and keep the `..._cli.py` file only to execute the command line instructions.

clinicadl/generate/artefacts/generate_trivial_noise_cli.py
```diff
@@ -0,0 +1,186 @@
...
+    load_and_check_tsv,
+    write_missing_mods,
+)
+from clinicadl.utils import cli_param
+from clinicadl.utils.caps_dataset.data import CapsDataset
+from clinicadl.utils.maps_manager.iotools import commandline_to_json
+
+logger = getLogger("clinicadl.generate")
+
+# def generate_noise_dataset(
```
ravih18: Why is this part commented ?

clinicadl/generate/generate_cli.py
```diff
@@ -24,7 +25,7 @@ def cli() -> None:
 cli.add_command(generate_trivial_cli)
 cli.add_command(generate_shepplogan_cli)
 cli.add_command(generate_hypo_cli)
-cli.add_command(generate_trivial_motion_cli)
+cli.add_command(generate_artifacts_cli)
```
ravih18, line 28: It looks like some new command line are missing : generate_trivial_motion/noise/contrast.

clinicadl/train_config.toml
```diff
@@ -0,0 +1,6 @@
+# CONFIG FILE FOR TESTS
```
ravih18: I am not sure this file should be there.

ravih18 (changes requested):

ravih18 (approved):

PR #465, opened 2023-08-17
Execute SSIM on GPU

ncassereau:
Currently, during validation, we compute SSIM (at least when we are doing reconstruction). But it's always executed on CPU. With this change, each SSIM computation checks if a GPU is available and uses it, making each validation step is roughly 5x faster.


camillebrianceau (approved):
Thanks for that, ready to merge for me! :)

ravih18 (approved):
Great, can be merged ! Indeed SSIM computation was bit longer than it should be.

PR #477, opened 2023-09-13
Add carbon emissions tracker callback

camillebrianceau:


pyproject.toml
```diff
@@ -28,20 +28,23 @@ classifiers = [
...
+clinica = "^0.7.6"
 torch = "^1.8.0"
 torchvision = "*"
 tensorboard = "*"
 toml = "*"
 pandas = "^1.2"
 numpy = "^1.17"
 scikit-learn = "^1.0"
-scikit-image = "^0.19"
+scikit-image = ">=0.21"
```
ghisvail: I'd advise keeping the caret operator. There is no guarantee the API won't be changed in a backward incompatible way in the future.

```suggestion
scikit-image = "^0.21"
```

clinicadl/utils/cli_param/train_option.py
```diff
@@ -304,3 +304,10 @@
     default=0,
     help="Number of layer that will be retrain during training. For example, if it is 2, the last two layers of the model will not be freezed.",
 )
+# informations
+emissions_tracker = cli_param.option_group.informations_group.option(
+    "--track_emissions/--dont_track_emissions",
```
ravih18: Is there the need to add a default ?
BTW the word `tracker` might be wrongly understood.

PR #497, opened 2023-11-08
Add others callbacks

camillebrianceau:


clinicadl/utils/callbacks/callbacks.py
```diff
@@ -122,3 +122,100 @@ def on_train_begin(self, parameters, **kwargs):
 
     def on_train_end(self, parameters, **kwargs):
         self.tracker.stop()
+
+
+class Tracker(Callback):
```
ncassereau, line 127: Something I should have corrected when introducing DDP is that MLflow is supposed to be running only on the master process, otherwise you're going to have duplicated logs. We could correct it in the fsdp PR, or this one, as you wish.

camillebrianceau: Can you do it in this PR  @ncassereau-idris ?

PR #505, opened 2023-12-05
minor changes 

camillebrianceau:


clinicadl/train/train_utils.py
```diff
@@ -21,6 +21,7 @@ def build_train_dict(config_file: Path, task: str) -> Dict[str, Any]:
     Returns:
         dictionary of values ready to use for the MapsManager
     """
+    print("test")
```
ncassereau: I assume it is meant to be removed.

clinicadl/utils/maps_manager/maps_manager.py
```diff
@@ -67,10 +67,10 @@ def __init__(
...
         """
         self.maps_path = maps_path.resolve()
-        if verbose is not None:
-            if verbose not in level_list:
-                raise ValueError(f"verbose value {verbose} must be in {level_list}.")
-            setup_logging(level_list.index(verbose))
+        # if verbose is not None:
+        #     if verbose not in level_list:
+        #         raise ValueError(f"verbose value {verbose} must be in {level_list}.")
+        #     setup_logging(level_list.index(verbose))
```
ncassereau: Wouldn't that break the logger ?  I assume that you're doing this because the maps_manager is instanciated multiple times and you want to avoid duplicating loggers. If you're doing this elsewhere, sorry I did not notice. Otherwise, perhaps you could avoid multiple instanciation somehow. For instance in #489 , I suggested that we could empty the logger handlers list (https://github.com/aramis-lab/clinicadl/blob/a1b5207a8b252e7fa99ab03a76725544d8efdf68/clinicadl/utils/logger.py#L58) so that we could instanciate the maps_manager multiple times safely. Or maybe click could do it once and for all, but I'm not sure how, I don't know this library well enough.

ncassereau (commented):
Just two other comments (not sure about their relevance):

- `get_subject_session_list` is imported but never used in generate.py. Not sure whether or not it supposed to be reintroduced later on.
- for the renaming of `remove_unique_session`. Is adding just an "_" clear enough ? Some convention do something like "should_remove_unique_session" or "remove_unique_session_fn". Idk if that's a good idea ^^

PR #507, opened 2023-12-08
Add option to save model for each epochs 

camillebrianceau:
PR to add an option to save the models from all other epochs, regardless of performance and not only the best-loss.

### DIFF
--- a/clinicadl/utils/cli_param/train_option.py
+++ b/clinicadl/utils/cli_param/train_option.py
@@ -82,11 +82,9 @@ compensation = cli_param.option_group.reproducibility_group.option(
     type=click.Choice(["memory", "time"]),
 )
 save_all_models = cli_param.option_group.reproducibility_group.option(
-    "--save_all_models",
+    "--save_all_models/--save_only_best_model",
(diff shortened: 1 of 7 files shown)


clinicadl/utils/cli_param/train_option.py
```diff
@@ -81,6 +81,14 @@
     # default="memory",
     type=click.Choice(["memory", "time"]),
 )
+save_all_models = cli_param.option_group.reproducibility_group.option(
+    "--save_all_models",
+    type=bool,
+    is_flag=True,
+    help="save gjseorimgsorg",
```
ncassereau: It's some form of elvish, I can't read it.

clinicadl/utils/maps_manager/maps_manager.py
```diff
@@ -2426,6 +2431,7 @@ def _write_weights(
         split: int,
         network: int = None,
         filename: str = "checkpoint.pth.tar",
+        save_all_models=False,
```
ncassereau: The type annotation is missing

PR #520, opened 2024-01-25
Add option for adaptive learning rate

camillebrianceau:
Add an adaptive learning rate with the function `ReduceLROnPlateau(optimizer, mode="min", factor=0.1, verbose=True )` . I didn't add any option to change the factors, the mode or the patience (default is 10). 
The update of the learning rate is based on the validation loss. 

Don't know if I need to add all these options or not.

### DIFF
--- a/clinicadl/utils/metric_module.py
+++ b/clinicadl/utils/metric_module.py
@@ -92,8 +92,6 @@ class MetricModule:
         Returns:
             (float) mean squared error
         """
-        print("le")
-        print("gjgor")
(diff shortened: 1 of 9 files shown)


clinicadl/utils/metric_module.py
```diff
@@ -92,7 +92,8 @@ def mse_fn(y, y_pred):
         Returns:
             (float) mean squared error
         """
-
+        print("le")
```
ncassereau: forgotten prints

ncassereau (approved):

PR #521, opened 2024-01-26
Extract tensor from BIDS

camillebrianceau:
Add `--from_bids` option in `prepare-data` pipeline. With the option, you need to provide a path to a BIDS from which the CAPS will be created. It creates a CAPS with only the tensors extracted to avoid any preprocessing. 

`clinicadl prepare-data image CAPS_DATASET t1-linear -ej extract_image.json --from_bids BIDS_DIRECTORY`


clinicadl/test.py
ncassereau, line 1: Did you intend on committing this file ? Maybe I'm wrong, but it looks like a mistake to me.

camillebrianceau:
change command line
 `clinicadl prepare-data-from-bids image BIDS_DIRECTORY CAPS_DIRECTORY MODALITY -ej extract_image.json`

PR #534, opened 2024-03-05
Use `setdefault()` method instead of looping over default dict

### LINKED ISSUE #533: Use `setdefault()` method instead of looping over default dict
This portion of the code can be rewritten in a simpler way using the `.setdefault()` method of Python dictionaries (for example  `space_dict.setdefault("d_reduction", "MaxPooling")`):

https://github.com/aramis-lab/clinicadl/blob/c6177e42a0ea7528b50a26864fe807d3f5e3cdb8/clinicadl/random_search/random_search_utils.py#L46-L57

thibaultdvx:
Fixes #533


NicolasGensollen (approved):
LGTM, thanks @thibaultdvx !

PR #535, opened 2024-03-05
Use parent or parents attributes of Pathlib Path object instead of chaining ".." strings

### LINKED ISSUE #532: Use `parent` or `parents` attributes of Pathlib Path object instead of chaining ".." strings
https://github.com/aramis-lab/clinicadl/blob/c6177e42a0ea7528b50a26864fe807d3f5e3cdb8/clinicadl/train/train_utils.py#L29

https://github.com/aramis-lab/clinicadl/blob/c6177e42a0ea7528b50a26864fe807d3f5e3cdb8/clinicadl/train/train_utils.py#L48

https://github.com/aramis-lab/clinicadl/blob/c6177e42a0ea7528b50a26864fe807d3f5e3cdb8/clinicadl/utils/maps_manager/maps_manager_utils.py#L22

thibaultdvx:
Fixes #532

### DIFF
--- a/clinicadl/hugging_face/hugging_face.py
+++ b/clinicadl/hugging_face/hugging_face.py
@@ -133,9 +133,7 @@ def create_readme(
         raise ClinicaDLArgumentError("There is no maps.json file in your repository.")
 
     clinicadl_root_dir = Path(__file__).parents[1]
-    config_path = (
-        Path(clinicadl_root_dir) / "resources" / "config" / "train_config.toml"
-    )
+    config_path = clinicadl_root_dir / "resources" / "config" / "train_config.toml"
     config_dict = toml.load(config_path)
 
     train_dict = read_json(config_file)
--- a/clinicadl/hugging_face/hugging_face.py
+++ b/clinicadl/hugging_face/hugging_face.py
@@ -132,7 +132,7 @@ def create_readme(
     if not config_file.is_file():
         raise ClinicaDLArgumentError("There is no maps.json file in your repository.")
 
-    clinicadl_root_dir = Path(__file__).parents[2]
+    clinicadl_root_dir = Path(__file__).parents[1]
     config_path = (
         Path(clinicadl_root_dir) / "resources" / "config" / "train_config.toml"
     )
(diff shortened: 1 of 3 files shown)


clinicadl/hugging_face/hugging_face.py
```diff
@@ -132,7 +132,7 @@ def create_readme(
     if not config_file.is_file():
         raise ClinicaDLArgumentError("There is no maps.json file in your repository.")
 
-    clinicadl_root_dir = (Path(__file__) / "../..").resolve()
+    clinicadl_root_dir = Path(__file__).parents[1]
```
NicolasGensollen, line 135: Even if not specified in the linked issue, I think we can also simplify the line just after this one.
Indeed, `clinical_root_dir` is already a Path object, such that there is no need to cast it.

clinicadl/train/train_utils.py
```diff
@@ -26,7 +26,7 @@ def build_train_dict(config_file: Path, task: str) -> Dict[str, Any]:
     """
     if config_file is None:
         # read default values
-        clinicadl_root_dir = (Path(__file__) / "../..").resolve()
+        clinicadl_root_dir = Path(__file__).parents[1]
```
NicolasGensollen, line 29: Same here

clinicadl/train/train_utils.py
```diff
@@ -45,7 +45,7 @@ def build_train_dict(config_file: Path, task: str) -> Dict[str, Any]:
             del user_dict["Random_Search"]
 
         # read default values
-        clinicadl_root_dir = (Path(__file__) / "../..").resolve()
+        clinicadl_root_dir = Path(__file__).parents[1]
```
NicolasGensollen, line 46: And here

NicolasGensollen (commented):
LGTM, thanks @thibaultdvx !
I think we can also make another small improvement while we are at it (see comments).

thibaultdvx:
Done @NicolasGensollen

NicolasGensollen (approved):
LGTM, thanks !

PR #541, opened 2024-03-18
[INFRA] Use python 3.11 in the readthedocs workflow

NicolasGensollen:
Something in the readthedocs pipeline seems to break with Poetry 1.8. 
This PR proposes to pin Poetry to 1.7 for building the docs until we find the issue with 1.8.
Also updates the Python version from 3.10 to 3.11.


camillebrianceau:
Thanks for that!

PR #544, opened 2024-03-19
fix typeerror 543

### LINKED ISSUE #543: TypeError: can only concatenate str (not "PosixPath") to str
When I run `clinicadl generate trivial` with `mask_path=None`, I get the following error: `TypeError: can only concatenate str (not "PosixPath") to str`, from:

https://github.com/aramis-lab/clinicadl/blob/076863c306726b3a2b9e23ff03c6ddbb2e6df1eb/clinicadl/generate/generate.py#L277-L280

It seems that `fetch_file` function returns a `PathLike` object rather than a `str`.

thibaultdvx:
Fixes #543

### DIFF
--- a/clinicadl/generate/generate.py
+++ b/clinicadl/generate/generate.py
@@ -277,7 +277,7 @@ def generate_trivial_dataset(
             try:
                 mask_path_tar = fetch_file(FILE1, cache_clinicadl)
                 tar_file = tarfile.open(mask_path_tar)
-                print("File: " + mask_path_tar)
+                print("File: " + str(mask_path_tar))
                 try:
                     tar_file.extractall(cache_clinicadl)
                     tar_file.close()
--- a/clinicadl/generate/generate.py
+++ b/clinicadl/generate/generate.py
@@ -277,7 +277,7 @@ def generate_trivial_dataset(
             try:
                 mask_path_tar = fetch_file(FILE1, cache_clinicadl)
                 tar_file = tarfile.open(mask_path_tar)
-                print("File: " + str(mask_path_tar))
+                print(f"File: {mask_path_tar}")
                 try:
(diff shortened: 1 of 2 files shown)


clinicadl/utils/clinica_utils.py
```diff
@@ -1146,12 +1146,12 @@ def fetch_file(remote: RemoteFileStructure, dirname: Optional[Path]) -> Path:
     remote : RemoteFileStructure
         Structure containing url, filename and checksum.
 
-    dirname : str
+    dirname : Path
```
thibaultdvx, line 1149: I'm not sure about `Optional[Path]` type hint for this parameter (see line 1141). If `dirname` is None, I feel like `dirname.exists()` (line 1158) will raise an error.

clinicadl/generate/generate.py
```diff
@@ -277,7 +277,7 @@ def generate_trivial_dataset(
             try:
                 mask_path_tar = fetch_file(FILE1, cache_clinicadl)
                 tar_file = tarfile.open(mask_path_tar)
-                print("File: " + mask_path_tar)
+                print("File: " + str(mask_path_tar))
```
NicolasGensollen:
```suggestion
                print(f"File: {mask_path_tar}")
```

clinicadl/utils/clinica_utils.py
NicolasGensollen: I agree, as of now, `dirname`cannot be None or things would break. I think we can update the type hint to `Path`.

NicolasGensollen (commented):
Thanks for fixing this @thibaultdvx !

NicolasGensollen (approved):
LGTM, thanks @thibaultdvx !

PR #545, opened 2024-03-19
fix BIDS file format for pet

### LINKED ISSUE #538: File format not BIDS compliant
https://github.com/aramis-lab/clinicadl/blob/ff2f892927a8304b171012486936ae38f3d9688c/clinicadl/utils/clinica_utils.py#L161

This file format is not BIDS compliant. Especially, `_pet_space` should be replaced with `_space`. It leads to an inconsistency between `ClinicaDL` and `Clinica`, which has already adopted this format:

https://github.com/aramis-lab/clinica/blob/82cf7649fd55c22e96ad8b6d043266ae3085eb63/clinica/pipelines/pet_linear/pet_linear_utils.py#L266-L275

thibaultdvx:
Fixes #538 

Edited by @NicolasGensollen :  Requires data PR: https://github.com/aramis-lab/clinicadl_data_ci/pull/34


NicolasGensollen (commented):
Thanks @thibaultdvx ! The fix is good.
Something that would be awesome is to add a unit test for this function.
You'd need to create a file `test/unittest/utils/test_clinica_utils.py` and add a test there. WDYT ?

thibaultdvx:
@NicolasGensollen thanks for the review. Yes I can do that!

I also checked why the CI checks failed and I think that it is because the files used to run the checks still have the old name format. Are these data stored on Inria's VMs? How can we access them?

tests/unittests/utils/test_clinica_utils.py
```diff
@@ -0,0 +1,29 @@
...
+        ] 
+
+)
+def test_pet_linear_nii(
+    acq_label, suvr_reference_region, uncropped_image, expected_pattern
+):
+    from clinicadl.utils.clinica_utils import pet_linear_nii
+
+    assert(
+        pet_linear_nii(acq_label, suvr_reference_region, uncropped_image)["pattern"] == expected_pattern
```
NicolasGensollen: I think we can test the full output while we are at it:

```suggestion
        pet_linear_nii(acq_label, suvr_reference_region, uncropped_image)["pattern"] == {
            "description": "",
            "needed_pipeline": "pet-linear",
            "pattern": expected_pattern,
        }
```

tests/unittests/utils/test_clinica_utils.py
```diff
@@ -0,0 +1,33 @@
...
+    ] 
+
+)
+def test_pet_linear_nii(
+    acq_label, suvr_reference_region, uncropped_image, expected_pattern
+):
+    from clinicadl.utils.clinica_utils import pet_linear_nii
+
+    assert(
+        pet_linear_nii(acq_label, suvr_reference_region, uncropped_image)["pattern"] == {
```
NicolasGensollen: Made a mistake in my previous suggestion:

```suggestion
        pet_linear_nii(acq_label, suvr_reference_region, uncropped_image) == {
```

NicolasGensollen:
@thibaultdvx At the moment the linter is not happy with your code (I made some recent changes to the infrastructure in #546 so there might be a difference between the one you're using and the one the CI is now using...).

What you can do is rebase on top of `dev` in order to get the changes from #546 (especially the pre-commit config). Then you should be able to install (or re-install) the pre-commit hook by running : 

```
pre-commit install
```

The linter, formatter, and spell checker will then run every time you make a commit (and they will be the same as the ones the CI is using).

Let me know if this is unclear or if you have questions.

camillebrianceau (approved):
LGTM ! thanks for that

PR #546, opened 2024-03-19
[INFRA] Use ruff instead of black and use codespell

NicolasGensollen:
This PR proposes to rely on [Ruff](https://docs.astral.sh/ruff/) rather than Black for linting.

Changes:

- modify the pre-commit hook to use Ruff (should be waaaay faster than black)
- modify the pyproject.toml to configure Ruff and Codespell (also remove previous config for black)
- modify the lint CI pipeline to rely on the pre-commit configuration
- restrict permissions to "read" for this pipeline

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -1,11 +1,14 @@
 name: 'Lint codebase'
+
 on:
   pull_request:
-    branches:
-      - dev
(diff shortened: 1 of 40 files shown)


camillebrianceau (approved):
LGTM ! Thanks for that @NicolasGensollen

PR #547, opened 2024-03-20
test to correct sorting with 2 digit

camillebrianceau:
Avec 2 digits, ClinicaDL ne pouvait pas trier les listes de (participants, sessions) si la liste contient des sessions > 100 (car 3 digits). 

Il faut changer la CI pour que les tests passent

### DIFF
--- a/clinicadl/tsvtools/get_labels/get_labels.py
+++ b/clinicadl/tsvtools/get_labels/get_labels.py
@@ -421,6 +421,6 @@ def get_labels(
     output_df.reset_index(inplace=True)
     output_df.sort_values(by=["participant_id", "session_index"], inplace=True)
     output_df.drop("session_index", axis=1, inplace=True)
-    output_df.to_csv(output_tsv, sep="\t")
+    output_df.to_csv(output_tsv, sep="\t", index=False)
 
     logger.info(f"Results are stored in {output_dir}.")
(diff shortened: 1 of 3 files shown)


clinicadl/tsvtools/get_labels/get_labels.py
```diff
@@ -404,14 +411,16 @@ def get_labels(
...
     output_df = diagnosis_removal(output_df, diagnoses)
     output_df = mod_selection(output_df, missing_mods_dict, modality)
     output_df = apply_restriction(output_df, restriction_path)
 
-    output_df.reset_index()
-    output_df.sort_values(by=["participant_id", "session_id"], inplace=True)
+    output_df.reset_index(inplace=True)
+    output_df.sort_values(by=["participant_id", "session_index"], inplace=True)
+    output_df.drop("session_index", axis=1, inplace=True)
     output_df.to_csv(output_tsv, sep="\t")
```
thibaultdvx: I had to change this line with `output_df.to_csv(output_tsv, sep="\t", index=False)` to pass the test. My hypothesis is that it worked before because `output_df.reset_index()` was useless because `inplace` was `False` and thus `participant_id`and `session_id` were still the indices.

thibaultdvx (changes requested):

clinicadl/tsvtools/get_labels/get_labels.py
camillebrianceau: Thanks for that !

PR #561, opened 2024-04-19
Trainer config

thibaultdvx:
1. Introduces config classes for the training pipeline. The base class is `BaseTaskConfig` and contains the parameters common to `classification`, `reconstruction` and `regression` (`base_training_config.py`). `ClassificationConfig`, `ReconstructionConfig` and `RegressionConfig` contain the parameters specific to the task. (`classification_config.py`, `regression_config.py` and `reconstruction_config.py`).

2. Introduces unit tests to test these config classes and their pydantic's validators (`test_base_training_config.py`, `test_classification_config.py`, `test_regression_config.py` and `test_reconstruction_config.py`).

3. Links the CLI to these config classes in order for the user to have access to the default values (`train_option.py`).

4. Modifies the `task_launcher` function so that it accepts config classes as input (`task_utils.py`, `classification_cli.py`, `regression_cli.py` and `reconstruction_cli.py`).

5. Renames and simplifies the `build_train_dict` function that reads TOML config files (`train_utils.py`) and changes accordingly the files where it is called (`random_search_utils.py`, `classification_cli.py`, `regression_cli.py` and `reconstruction_cli.py`).

6. Adds a unit test for `extract_config_from_toml_file` (the old `build_train_dict` function) (`test_train_utils.py`).

7. Minor changes in the config TOML file template (`train_config.toml`).

Unrelated:

8. Get trainer's folder out of MAPSManager's folder.


clinicadl/train/tasks/base_training_config.py
```diff
@@ -0,0 +1,144 @@
...
+    tsv_directory: Path
+    output_maps_directory: Path
+    # Computational
+    gpu: bool = True
+    n_proc: int = 2
+    batch_size: int = 8
+    evaluation_steps: int = 0
+    fully_sharded_data_parallel: bool = False
+    amp: bool = False
+    # Reproducibility
```
NicolasGensollen, line 66: The number of attributes of this class is quite large. So large you felt it was necessary to use comments to organize them. Do you think it would be possible to split this class in multiple config classes ?

Something that would look more or less like this:

```python
class BaseTaskConfig(BaseModel):
    ...
    computational_config = ComputationalConfig()
    reproducibility_config = ReproducibilityConfig()
    ...

class ComputationalConfig(BaseModel):
    gpu: bool = True
    n_proc: int = 2
    ...
```

clinicadl/train/tasks/base_training_config.py
```diff
@@ -0,0 +1,144 @@
...
+    Base class to handle parameters of the training pipeline.
+    """
+
+    caps_directory: Path
+    preprocessing_json: Path
+    tsv_directory: Path
+    output_maps_directory: Path
+    # Computational
+    gpu: bool = True
+    n_proc: int = 2
```
NicolasGensollen, line 61: Why 2 and not 1 ? 🤔

clinicadl/train/tasks/base_training_config.py
```diff
@@ -0,0 +1,144 @@
...
+    batch_size: int = 8
+    evaluation_steps: int = 0
+    fully_sharded_data_parallel: bool = False
+    amp: bool = False
+    # Reproducibility
+    seed: int = 0
+    deterministic: bool = False
+    compensation: Literal["memory", "time"] = "memory"
```
NicolasGensollen: It could be useful to define proper types for these variables that have a finite number of possible values. Using enumerations can be a good strategy for that:

```python
...
compensation: Compensation = Compensation.MEMORY
...

class Compensation(str, Enum):
    """Possible compensations in clinicaDL."""

    MEMORY = "memory"
    TIME = "time"
```

WDYT ?

clinicadl/train/tasks/base_training_config.py
```diff
@@ -0,0 +1,144 @@
...
+    # Data
+    multi_cohort: bool = False
+    diagnoses: Tuple[str, ...] = ("AD", "CN")
+    baseline: bool = False
+    valid_longitudinal: bool = False
+    normalize: bool = True
+    data_augmentation: Tuple[str, ...] = ()
+    sampler: Literal["random", "weighted"] = "random"
+    size_reduction: bool = False
+    size_reduction_factor: Literal[2, 3, 4, 5] = 2
```
NicolasGensollen: Maybe I misunderstand the logic that will be implemented, but it seems like you could have a single attribute for that:

```python
size_reduction_factor: Optional[SizeReductionFactor] = None
```

with

```python
class SizeReductionFactor(int, Enum):
    """Available size reduction factors in ClinicaDL."""

    TWO = 2
    THREE = 3
    FOUR = 4
    FIVE = 5
```

And the previous bool attribute is equivalent to the factor being None:

```python
if size_reduction_factor:
    # do something with it
``` 

WDYT ?

clinicadl/train/tasks/base_training_config.py
```diff
@@ -0,0 +1,144 @@
...
+    caps_target: Path = Path("")
+    tsv_target_lab: Path = Path("")
+    tsv_target_unlab: Path = Path("")
+    preprocessing_dict_target: Path = Path(
+        ""
+    )  # TODO : change name in commandline. preprocessing_json_target?
+    # Cross validation
+    n_splits: int = 0
```
NicolasGensollen, line 94: Couldn't `n_splits` be defined as a property?

```python
@property
def n_splits(self):
    return len(self.split)
```

clinicadl/train/tasks/base_training_config.py
```diff
@@ -0,0 +1,144 @@
...
+        available_optimizers = cls.get_available_optimizers()
+        assert (
+            v in available_optimizers
+        ), f"Optimizer '{v}' not supported. Please choose among: {available_optimizers}"
+        return v
+
+    @classmethod
+    def get_available_transforms(cls) -> List[str]:
+        """To get the list of available transforms."""
+        available_transforms = [  # TODO : connect to transforms module
```
NicolasGensollen, line 163: It feels like there will be a transforms module at some point with a bunch of public classes defined in it and probably a factory method for conveniently importing an object from its name:

```python
class TranformNames(str, Enum):
    NOISE = "noise"
    ERASING = "erasing"
    ...

def transform_factory(name: Union[str, TransformName]) -> BaseTransform:
    name = TransformName(name) # this will manage the string input validation for you
    if name == TransformName.NOISE:
        return NoiseTransform
    if name == TransformName.ERASING:
        return ErasingTransform
    ...
```

All of that to say that the factory of the transforms module should handle the input validation. There is no need to have it here. These lists of possible values hidden deep in large classes are very easy to forget when adding new options for example.

clinicadl/train/tasks/reconstruction_config.py
```diff
@@ -0,0 +1,52 @@
...
+    @field_validator("selection_metrics", mode="before")
+    def list_to_tuples(cls, v):
+        if isinstance(v, list):
+            return tuple(v)
+        return v
+
+    @classmethod
+    def get_compatible_losses(cls) -> List[str]:
+        """To get the list of losses implemented and compatible with this task."""
+        compatible_losses = [  # TODO : connect to the Loss module
```
NicolasGensollen, line 46: The previous pattern could be used in different spots like here.

NicolasGensollen (commented):
Hi @thibaultdvx 

I was taking a look at the different improvements PR the clinicaDL team is currently working on.
Even though this is still in draft mode, I made a few early suggestions along the way, mostly on how things could be organized.
Since I haven't followed the sprint I might be missing important points, so feel free to ignore.

clinicadl/train/tasks/base_training_config.py
thibaultdvx: Hi @NicolasGensollen, thanks for contributing! Indeed, this class is quite complex because it contains all the parameters of the training pipeline, but it is temporary. We will get attributes out of it in the next steps of the refactoring.

clinicadl/train/tasks/base_training_config.py
thibaultdvx: Good question... I kept the default values as they were in ClinicaDL

clinicadl/train/tasks/base_training_config.py
thibaultdvx: Interesting, I hadn't thought of that. What are the benefits of this approach?

clinicadl/train/tasks/base_training_config.py
thibaultdvx: Yes you're right. It seems that `size_reduction` and `size_reduction_factor` are quite redundant. 

It requires to make changes in functions that I have not modified yet, and I feel that this PR is already big enough. So, I'll keep it in mind for the next one!

clinicadl/train/tasks/base_training_config.py
NicolasGensollen: Alright ! Sounds like a good plan !

clinicadl/train/tasks/base_training_config.py
NicolasGensollen: Python being what it is, you can put type hints on a class or function and give any kind of value at runtime and you wouldn't get any error. For example this is totally possible at runtime:

```python
>>> from typing import Literal
>>> class Config:
...     compensation: Literal["foo", "bar"] = "foo"
...
>>> c = Config()
>>> c.compensation
'foo'
>>> c.compensation = "toto"
>>> c.compensation
'toto'
```

This can potentially cause nasty bugs later, somewhere else in the code. If you want to have a more strict control over that, you need to work a little harder and do something like:

```python
>>> class Config:
...     _compensation: Compensation = Compensation.FOO
...     @property
...     def compensation(self) -> Compensation:
...         return self._compensation
...     @compensation.setter
...     def compensation(self, value: str | Compensation):
...         self._compensation = Compensation(value)
>>> c = Config()
>>> c.compensation
<Compensation.FOO: 'foo'>
>>> c.compensation = "toto"
  File "<stdin>", line 1, in <module>
  File "<stdin>", line 8, in compensation
  File "/Users/nicolas.gensollen/opt/anaconda3/envs/clinica-t/lib/python3.11/enum.py", line 712, in __call__
    return cls.__new__(cls, value)
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/nicolas.gensollen/opt/anaconda3/envs/clinica-t/lib/python3.11/enum.py", line 1135, in __new__
    raise ve_exc
ValueError: 'toto' is not a valid Compensation
>>> c.compensation = "bar"
>>> c.compensation
<Compensation.BAR: 'bar'>
```

The enumeration provides both the validation logic as well as a richer type than a plain string.

clinicadl/train/tasks/base_training_config.py
NicolasGensollen: Sounds good 👍

clinicadl/train/tasks/base_training_config.py
```diff
@@ -0,0 +1,149 @@
...
+    data_augmentation: Tuple[str, ...] = ()
+    sampler: Literal["random", "weighted"] = "random"
+    size_reduction: bool = False
+    size_reduction_factor: Literal[2, 3, 4, 5] = 2
+    caps_target: Path = Path("")
+    tsv_target_lab: Path = Path("")
+    tsv_target_unlab: Path = Path("")
+    preprocessing_dict_target: Path = Path(
+        ""
+    )  # TODO : change name in commandline. preprocessing_json_target?
```
camillebrianceau: I think that all the `target` options are for SSDA and multi-cohort/modalities so we'll change everything in `CapsDataset` PR

clinicadl/train/tasks/base_training_config.py
camillebrianceau: I think that `n_splits` is the total number of splits and `split` is the ones you want to train on

clinicadl/train/tasks/base_training_config.py
```diff
@@ -0,0 +1,149 @@
...
+        validate_assignment = True
+
+    @field_validator("diagnoses", "split", "data_augmentation", mode="before")
+    def list_to_tuples(cls, v):
+        if isinstance(v, list):
+            return tuple(v)
+        return v
+
+    @field_validator("transfer_path", mode="before")
+    def false_to_none(cls, v):
```
camillebrianceau, line 130: Why do you have to create this method? Is it not already `False`?

clinicadl/train/train_utils.py
```diff
@@ -14,72 +14,70 @@
...
         raise ClinicaDLConfigurationError(
-            f"config_file {config_file} should be a TOML or a JSON file."
+            f"Config file {config_file} should be a TOML file."
         )
+
+    user_dict = toml.load(config_file)
+    if "Random_Search" in user_dict:
+        del user_dict["Random_Search"]
+
+    # get the template
```
camillebrianceau, line 53: For me, this part is not needed anymore. With the config class, we don't need the `train_config.toml` file because default values are already in the config class so we just need to instantiate a `TrainConfig` and add a `get_parameters_from_toml` method or something like that. WDYT?

clinicadl/utils/cli_param/train_option.py
```diff
@@ -1,390 +1,432 @@
...
+
 import click
 
+from clinicadl.train.tasks.base_training_config import BaseTaskConfig
+from clinicadl.train.tasks.classification_config import ClassificationConfig
+from clinicadl.train.tasks.reconstruction_config import ReconstructionConfig
+from clinicadl.train.tasks.regression_config import RegressionConfig
 from clinicadl.utils import cli_param
 
+# Arguments
```
camillebrianceau, line 11: Can't we set the type of arguments from the `BaseTaskConfig` like it's done after for the options?

clinicadl/utils/cli_param/train_option.py
```diff
@@ -1,390 +1,432 @@
...
+# Config file
 config_file = click.option(
     "--config_file",
     "-c",
     type=click.Path(exists=True),
     help="Path to the TOML or JSON file containing the values of the options needed for training.",
 )
+
+# Options #
+base_config = BaseTaskConfig
```
camillebrianceau: Maybe you can do ` base_config = BaseTaskConfig.model_fields`so you don't have to call the `model_fields`method every time ?

clinicadl/utils/cli_param/train_option.py
```diff
@@ -1,390 +1,432 @@
...
+reconstruction_config = ReconstructionConfig
+
 # Computational
 gpu = cli_param.option_group.computational_group.option(
     "--gpu/--no-gpu",
-    type=bool,
-    default=None,
+    default=base_config.model_fields["gpu"].default,
     help="Use GPU by default. Please specify `--no-gpu` to force using CPU.",
+    show_default=True,
```
camillebrianceau, line 38: Why the `show_default=True` is needed?

clinicadl/utils/cli_param/train_option.py
```diff
@@ -1,390 +1,432 @@
...
 fully_sharded_data_parallel = cli_param.option_group.computational_group.option(
     "--fully_sharded_data_parallel",
     "-fsdp",
-    type=bool,
     is_flag=True,
     help="Enables Fully Sharded Data Parallel with Pytorch to save memory at the cost of communications. "
     "Currently this only enables ZeRO Stage 1 but will be entirely replaced by FSDP in a later patch, "
     "this flag is already set to FSDP to that the zero flag is never actually removed.",
-    default=False,
 )
```
camillebrianceau, line 71: no type or default ?

clinicadl/utils/cli_param/train_option.py
```diff
@@ -1,390 +1,432 @@
...
-
 amp = cli_param.option_group.computational_group.option(
     "--amp/--no-amp",
-    type=bool,
+    default=base_config.model_fields["amp"].default,
```
camillebrianceau: no type ?

clinicadl/utils/cli_param/train_option.py
camillebrianceau, line 1: I think sometimes `type` or `default` is missing. Not sure if it is on purpose or not?

clinicadl/utils/maps_manager/maps_manager.py
```diff
@@ -529,6 +529,12 @@ def write_parameters(json_path: Path, parameters, verbose=True):
         if verbose:
             logger.info(f"Path of json file: {json_path}")
 
+        # temporary: to match CLI data. TODO : change CLI data
```
camillebrianceau, line 532: I think it can be done in the `path_encoder` called in `json.dump`.  Maybe in the next PR because the `path_encoder` and `path_decoder` need real refactoring but to keep in mind

camillebrianceau (commented):
Some minor comments, feel free to ignore :)

PR #565, opened 2024-04-25
Put train pipeline inside CLI files

thibaultdvx:
1. Removes the `train` function to directly call `MAPSManager` and `Trainer` it in the CLI files. And deletes `train.py` that was then empty.

2. Renames `task_launcher` to `preprocessing_json_reader` (because of 1., the only purpose of `task_launcher` was then to read preprocessing files).

3. Puts `preprocessing_json_reader` in `train_utils.py` and deletes `task_utils.py` that was then empty.

4. Creates a `merge_cli_and_config_file_options` (in `train_utils.py`) to merge the options the user gave via the CLI with those found in the TOML config file.

5. Creates unit tests for `preprocessing_json_reader` and `merge_cli_and_config_file_options`.

For the next PRs:

- Modify `MAPSManager`and `Trainer` so that they accept a config object.
- Reorganize `train` folder and put the `trainer` folder inside. 
- I'm not really satisfy with having to call `preprocessing_json_reader` because it means that the user will have to do that in the API mode. `preprocessing_json_reader` could maybe be put in `BaseTaskConfig` and called in `__init__`.


clinicadl/train/train_utils.py
```diff
@@ -155,3 +157,137 @@ def get_model_list(architecture=None, input_size=None, model_layers=False):
...
+            if not json_found:
+                raise ValueError(
+                    f"Preprocessing JSON {preprocessing_json_target} was not found for any CAPS "
+                    f"in {caps_target}."
+                )
+
+    # Mode and preprocessing
+    preprocessing_dict = read_preprocessing(preprocessing_json)
+    config._preprocessing_dict = preprocessing_dict
+    config._mode = preprocessing_dict["mode"]
```
NicolasGensollen, line 242: This is something to keep an eye on. You're mutating attributes of this `BaseTaskConfig` class which are supposed to be private. You might want to implement getters and setters if some safety checks are needed, or make them public depending on what you want to do.

thibaultdvx: @NicolasGensollen yes you're right, it is a bit weird for the moment. Actually, I put them as private attribute because they are parameters that shouldn't be saved in MAPS at the end. Anyways, I'm not a fan of this `preprocessing_json_reader` function, so it will probably be modified.

As for safety checks, there is indeed no checks made on private attributes. So, I'll note to add them in a next PR.

camillebrianceau (approved):
LGTM !

PR #567, opened 2024-04-29
Clean train cli

thibaultdvx:
1. Creates Enum classes for Transforms, Optimizers, Metrics and Losses.
2. Creates an enum file to gather enum classes.
3. Creates generic `get_default_from_config_class` and `get_type_from_config_class` (in `clinicadl/utils/config_utils.py`) to get types and defaults of parameters in config classes.
4. Creates unit tests for these two functions.
5. Uses these two functions in CLI files.
6. Adds PositiveFloat, PositiveInt, NonNegativeFloat and NonNegativeInt typing in config classes.


camillebrianceau (approved):
LGTM !

PR #568, opened 2024-04-29
Creation of Prepare Data Config

camillebrianceau:
- Creation of the `transforms` folder and files
- Creation of the `PrepareDataConfig` and so the `PrepareDataImageConfig`, `PrepareDataPatchConfig`, `PrepareDataSliceConfig` and `PrepareDataRoiConfig`
- Creation of the `prepare_data_param` folder with the CLI adapted to the Config classes
- Creation of new Enum: `DTIMeasure`, `DTISpace`, `ExtractionMethod`, `SliceDirection`, `SliceMode`,  `Template`, `Pattern` 
- Adaptation of the code and the tests to these changes. 

Future work: creation of unittests, better adaptation to Config in the data.py file,


clinicadl/utils/caps_dataset/data.py
```diff
@@ -416,6 +429,14 @@ def __init__(
             transformations=all_transformations,
             multi_cohort=multi_cohort,
         )
+        self.config = PrepareDataPatchConfig(
+            caps_directory=caps_directory,
+            preprocessing_cls=Preprocessing(preprocessing_dict["preprocessing"]),
+            use_uncropped_image=preprocessing_dict["use_uncropped_image"],
+            save_features=preprocessing_dict["prepare_dl"],
+            patch_size=preprocessing_dict["patch_size"],
+            stride_size=preprocessing_dict["stride_size"],
```
NicolasGensollen, line 450: Something that could be useful is to define custom constructors for these classes:

```python
class PrepareDataPatchConfig:
    caps_directory: Path
    preprocessing_cls: Preprocessing
    use_uncropped_image: ...

    @classmethod
    def from_preprocessing(cls, caps_directory: Path, preprocessing: dict):
        return cls(
            caps_directory,
            Preprocessing(preprocessing_dict["preprocessing"]),
            preprocessing_dict["use_uncropped_image"],
            ....
        )
```

This way, the code would be very simple:

```python
self.config = PrepareDataPatchConfig.from_preprocessing(caps_directory, preprocessing_dict)
```

clinicadl/utils/caps_dataset/data.py
```diff
@@ -526,6 +547,15 @@ def __init__(
             multi_cohort=multi_cohort,
         )
 
+        self.config = PrepareDataROIConfig(
```
NicolasGensollen, line 562: You might be able to use inheritance between these classes (if it makes sense) to factorize some code.

clinicadl/utils/caps_dataset/data.py
camillebrianceau: But these lines are just temporary...

clinicadl/prepare_data/prepare_data_cli.py
```diff
@@ -3,317 +3,176 @@
...
-    n_proc: int,
-    subjects_sessions_tsv: Optional[Path] = None,
-    extract_json: Optional[str] = None,
-    use_uncropped_image: bool = False,
-    tracer: Optional[str] = None,
-    suvr_reference_region: Optional[str] = None,
-    custom_suffix: str = "",
-    dti_measure: str = "FA",
-    dti_space: str = "*",
+    preprocessing: Preprocessing,
```
thibaultdvx, line 41: Tell me if i'm wrong but I think that `preprocessing` will be a `str` because it comes from the command line. I would thus suggest to change the type hint to `str`, and change the type of preprocessing at the beginning of the function with `preprocessing = Preprocessing(preprocessing)`. WDYT?

clinicadl/prepare_data/prepare_data_cli.py
```diff
@@ -3,317 +3,176 @@
...
+@prepare_data_param.option.extract_json
+@prepare_data_param.option.use_uncropped_image
+@prepare_data_param.option.tracer
+@prepare_data_param.option.suvr_reference_region
+@prepare_data_param.option.custom_suffix
+@prepare_data_param.option.dti_measure
+@prepare_data_param.option.dti_space
+@prepare_data_param.option_patch.patch_size
+@prepare_data_param.option_patch.stride_size
+def patch_cli(caps_directory: Path, preprocessing: Preprocessing, **kwargs):
```
thibaultdvx, line 79: Same remark with `preprocessing`'s type.

clinicadl/prepare_data/prepare_data_cli.py
```diff
@@ -3,317 +3,176 @@
...
-@cli_param.option.subjects_sessions_tsv
-@cli_param.option.extract_json
-@cli_param.option.use_uncropped_image
-@cli_param.option.tracer
-@cli_param.option.suvr_reference_region
-@cli_param.option.custom_suffix
-@cli_param.option.dti_measure
-@cli_param.option.dti_space
+@prepare_data_param.argument.caps_directory
+@prepare_data_param.argument.preprocessing
```
thibaultdvx, line 29: So `preprocessing` replaces `modality`? Then, the docstrings of the following cli functions (e.g. `image_cli`) should be changed. And we must keep in mind that the documentation should be changed as well.

clinicadl/prepare_data/prepare_data_config.py
```diff
@@ -0,0 +1,134 @@
...
+    suvr_reference_region_cls: SUVRReferenceRegions = (
+        SUVRReferenceRegions.CEREBELLUMPONS2
+    )
+    custom_suffix: str = ""
+    dti_measure_cls: DTIMeasure = DTIMeasure.FRACTIONAL_ANISOTROPY
+    dti_space_cls: DTISpace = DTISpace.ALL
+    save_features: bool = False
+    extract_method: ExtractionMethod
+
+    class ConfigDict:
```
thibaultdvx: Sorry @camillebrianceau, I told that you that we should use `ConfigDict` instead of `Config` (which is deprecated), but it turns out that they are not use in the same way. Some examples:

```python
class ClassTest(BaseModel):
    a: str = "a"
    class Config:
        validate_assignment=True

test = ClassTest()
test.a = 0
```
will raise a `ValidationError` as expected, but  the `Config` class usage is deprecated.

```python
class ClassTest(BaseModel):
    a: str = "a"
    class ConfigDict:
        validate_assignment=True

test = ClassTest()
test.a = 0
```
will not raise any error unfortunately.

```python 
from pydantic import BaseModel, ConfigDict
class ClassTest(BaseModel):
    a: str = "a"
    model_config = ConfigDict(validate_assignment=True)

test = ClassTest()
test.a = 0
```
will raise a `ValidationError` as expected.

To sum up, we should use the last option.

clinicadl/prepare_data/prepare_data_param/option.py
```diff
@@ -0,0 +1,104 @@
...
+    Tracer,
+)
+
+config = PrepareDataConfig.model_fields
+
+n_proc = click.option(
+    "-np",
+    "--n_proc",
+    type=config["n_proc"].annotation,
+    default=config["n_proc"].default,
```
thibaultdvx, line 21: Since PR #567, there are functions `get_default_from_config_class` and `get_type_from_config_class` in `clinicadl/utils/config_utils.py`, if you want to use them ;)

clinicadl/prepare_data/prepare_data_param/option.py
```diff
@@ -0,0 +1,104 @@
...
+    ),
+    show_default=True,
+)
+custom_suffix = click.option(
+    "-cn",
+    "--custom_suffix",
+    type=config["custom_suffix"].annotation,
+    default=config["custom_suffix"].default,
+    help=(
+        "Suffix of output files if MODALITY is `custom`. "
```
thibaultdvx: `MODALITY` is now `PREPROCESSING`?

thibaultdvx (commented):
LGTM, except a mistake i told you about `ConfigDict`. Plus some minor remarks

clinicadl/prepare_data/prepare_data_cli.py
camillebrianceau: Here yes, because we need the type of preprocessing used. Modality and Preprocessing were the same options. I changed the docstring!

clinicadl/prepare_data/prepare_data_param/option.py
camillebrianceau: I'll change that in the next PR!

PR #572, opened 2024-05-14
BUG in CI tests for prepare_data

camillebrianceau:
compare folder function wasn't well indented in the test_prepare_data.py file so we missed some bug


camillebrianceau:
Finally, I'm not sure if we have to take the `assert` out of the loop. I think tests need to be changed but not this way.

PR #573, opened 2024-05-15
[CI] Run tests through GitHub Actions

NicolasGensollen:
This PR proposes to run the tests through GitHub Actions rather than Jenkins (similar to what we did on Clinica a few months ago).

The VM that was used until now is running on CentOS7 and was clearly missing recent updates.
In addition, CentOS7 will be EOL in June 2024 (https://endoflife.date/centos).
Because of these reasons, I created and configured a new VM running on Ubuntu 20.04-lts.
I should have installed the required tools that we use (like dvc, pipx, poetry, miniconda...) but let me know if something is missing.

I also deployed a new GitHub runner on this VM (there is also one deployed on the CentOS VM but the workflows are crashing due to an old version of glibc, I'll delete it later).

This PR proposes to port two tests: `test_cli.py` and `test_tsvtools.py` to Github Actions, making sure that tests with and without access to `clinical_data_ci` are working fine.

If the proposed solution is accepted, I'll port the remaining tests and remove the associated Jenkins pipeline.

I will also need to take a closer look at the machine were are using for testing GPU support to see if we could implement something similar.

### DIFF
--- a/.github/workflows/test_cli.yml
+++ b/.github/workflows/test_cli.yml
@@ -32,7 +32,6 @@ jobs:
       - uses: actions/setup-python@v5
         with:
           python-version: ${{ env.PYTHON_VERSION }}
-          cache: poetry
       - name: Run CLI tests
         run: |
           make env.conda
--- a/.github/workflows/test_cli.yml
+++ b/.github/workflows/test_cli.yml
@@ -1,0 +1,46 @@
+name: CLI Tests
+
+on:
+  push:
+    branches: ["dev"]
+  pull_request:
+    branches: ["dev"]
+
+permissions:
+  contents: read
+
+concurrency:
+  group: '${{ github.workflow }} @ ${{ github.event.pull_request.head.label || github.head_ref || github.ref }}'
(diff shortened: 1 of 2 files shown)


NicolasGensollen:
Here is the result for test_cli: https://github.com/aramis-lab/clinicadl/actions/runs/9111679617/job/25049309668?pr=573

And the one for test_tsvtools: https://github.com/aramis-lab/clinicadl/actions/runs/9111679640/job/25049309674?pr=573

thibaultdvx (approved):
LGTM @NicolasGensollen, thanks a lot! I can take care of the other tests if you want.

NicolasGensollen:
Thanks for the review @thibaultdvx ! 👍 
I'll take care of the other tests no worries. I still need to deploy runners on the GPU machine (and maybe an aramis workstation), so I don't mind finishing.
@camillebrianceau is this good for you or do you want to take a closer look before we merge ?

NicolasGensollen:
I'm gonna go ahead and merge this.
We can always revert if we don't like it in the end...

PR #580, opened 2024-05-22
Create config folder

camillebrianceau:
In this PR, I created 2 folders for Clinicadl parameters configuration: 
- `config/config` contains a file for each DataConfig class
- `config/options` contains a file for the options corresponding to the data class of the same name.


clinicadl/config/config/callbacks.py
```diff
@@ -0,0 +1,22 @@
...
+from enum import Enum
+from logging import getLogger
+from pathlib import Path
+from typing import Any, Dict, Optional, Tuple
+
+from pydantic import BaseModel, ConfigDict, computed_field, field_validator
+from pydantic.types import NonNegativeFloat, NonNegativeInt, PositiveFloat, PositiveInt
+
+from clinicadl.utils.enum import ExperimentTracking
+from clinicadl.utils.preprocessing import read_preprocessing
```
thibaultdvx, line 11: We don't need all those imports, do we?

thibaultdvx (commented):
LGTM @camillebrianceau, thanks for putting some order!

clinicadl/config/config/callbacks.py
camillebrianceau: No! I'll change that in the next PR!

PR #581, opened 2024-05-22
[CI] Port remaining non GPU tests to GitHub Actions

NicolasGensollen:
Following #573 this PR proposes to port the remaining tests of the non-GPU Jenkins pipeline to Github Actions.
I kept the Jenkins file as is and will remove the pipeline in a follow-up PR if all went well.

### DIFF
--- a/.github/workflows/test_generate.yml
+++ b/.github/workflows/test_generate.yml
@@ -1,0 +1,51 @@
+name: Generate Tests
+
+on:
+  push:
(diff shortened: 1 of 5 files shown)


NicolasGensollen:
I implemented the symlink hack mentioned in #591 to make the tests pass while we decide what to do.

thibaultdvx (approved):

PR #583, opened 2024-05-22
generate pipeline adaptation to data classes

camillebrianceau:
Changes in this PR: 
- Creation of a `caps_dataset` folder which will be the main folder for all data-related files
- creation of `create_caps_dataset_config` function that returns the CapsDatsetConfig with the right modality and preprocessing
- move `return_dataset` function to a `data_utils.py` file
- made some changes in generate files to facilitate the creation of new functions for API mode
- creation of `get_mask_path` function
- change options of generate pipelines files 
- move some config files remaining in the config folder where they belong (interpret, predict and train)
- add `GenerateType` and `MaskChecksum` enums
- add a `read_utils.py` file (maybe to move to the new maps_manager after ???)
- minor changes like:
    -  `nib.load`-> `nib.loadsave.load`
    -  `nib.Nifti1`-> `nib.nifti1.Nifti1`
    -  `data_df.loc`-> `data_df.at` when needed


clinicadl/generate/generate_config.py
```diff
@@ -106,17 +82,9 @@ def gamma_validator(cls, v):
 
 class GenerateHypometabolicConfig(SharedGenerateConfigOne):
     anomaly_degree: float = 30.0
-    pathology_cls: Pathology = Pathology.AD
+    pathology: Pathology = Pathology.AD
     sigma: int = 5
```
thibaultdvx:
```suggestion
    sigma: PositiveFloat = 5.0
```

clinicadl/generate/generate_config.py
```diff
@@ -48,45 +48,21 @@ def check_tsv_file(cls, v):
...
     contrast: bool = False
-    gamma: Annotated[list[float], 2] = [-0.2, -0.05]
+    gamma: Tuple[float, float] = (-0.2, -0.05)
     motion: bool = False
     num_transforms: int = 2
     noise: bool = False
-    noise_std: Annotated[list[float], 2] = [5, 15]
-    rotation: Annotated[list[float], 2] = [2, 4]  # float o int ???
-    translation: Annotated[list[float], 2] = [2, 4]
+    noise_std: Tuple[float, float] = (5, 15)
```
thibaultdvx:
```suggestion
    noise_std: Tuple[PositiveFloat, PositiveFloat] = (5.0, 15.0)
```

thibaultdvx (approved):
LGTM @camillebrianceau!

A remark: it is obviously not the point of your PR, but I would suggest to precise the types in a next PR. With `pydantic`, we can access interesting types like `PositiveFloat` or  `NonNegativeFloat` (and their `int` equivalents).
It takes some reflexion to decide the right type (for example we need to ask ourselves: can this parameter be equal to zero?), but I think the more precise we are, the better.
WDYT?

clinicadl/generate/generate_config.py
thibaultdvx: These are examples to show what I meant in my remark. Feel free to ignore in this PR ;)

clinicadl/caps_dataset/caps_dataset_config.py
```diff
@@ -80,14 +89,11 @@ def create_caps_dataset_config(
...
-        get_preprocessing(extract_method),
-    ):
-        # pydantic config
-        model_config = ConfigDict(
-            validate_assignment=True, arbitrary_types_allowed=True
-        )
+    class CapsDatasetConfig(CapsDatasetBase):
+        modality: get_modality(preprocessing_type)
+        preprocessing: get_preprocessing(extract_method)
+
```
NicolasGensollen: I'm jumping in the middle of this, so I might be missing things, but I think having this class defined inside a function is really strange. Looking at this piece of code, it feels like you need a factory or a special constructor to instantiate the class `CapsDatasetConfig` depending on the preprocessing and the extraction method provided.

So, I think something more or less like this should do the trick:

```python
class CapsDatasetConfig(CapsDatasetBase):
    @classmethod
    def from_preprocessing_and_extraction_method(
        cls,
        data: DataConfig,
        preprocessing:  Union[str, Preprocessing],
        extraction: Union[str, ExtractionMethod],
    ):
        return cls(
            data,
            get_modality(Preprocessing(preprocessing)),
            get_preprocessing(ExtractionMethod(extraction)),
        )
```

You'd then use it like this:

```python
config = CapsDatasetConfig.from_preprocessing_and_extraction_method(
    data, "my_preproc", "my_extraction"
)
```

Does this make sense ?

clinicadl/caps_dataset/data_config.py
```diff
@@ -65,6 +79,54 @@ def check_label(self, _label: str):
         if not self.label:
             self.label = _label
 
+    @field_validator("mask_path", mode="before")
+    @classmethod
+    def check_mask_path(cls, v) -> Path:
+        if v is not None and not isinstance(v, Path):
+            Path(v)
+        return v
```
thibaultdvx: I'm not sure to understand the goal of this validator. I think that Pydantic will try to put the value passed to `mask_path` in a `Path` anyways:

```python
>>> from pydantic import BaseModel
>>> from typing import Optional
>>> from pathlib import Path
>>> class DataConfig(BaseModel):
...     mask_path: Optional[Path] = None
... 
>>> DataConfig(mask_path="path")
DataConfig(mask_path=PosixPath('path'))
>>> DataConfig(mask_path=0)
  File "<stdin>", line 1, in <module>
  File "/Users/thibault.devarax/anaconda3/envs/clinicadl_dev/lib/python3.9/site-packages/pydantic/main.py", line 175, in __init__
    self.__pydantic_validator__.validate_python(data, self_instance=self)
pydantic_core._pydantic_core.ValidationError: 1 validation error for DataConfig
mask_path
  Input is not a valid path [type=path_type, input_value=0, input_type=int]
```
So I feel that this validator is redundant with Pydantic's validators.

clinicadl/caps_dataset/data_config.py
```diff
@@ -65,6 +79,54 @@ def check_label(self, _label: str):
...
+
+    @field_validator("data_tsv", mode="before")
+    @classmethod
+    def check_data_tsv(cls, v) -> Path:
+        if v is not None:
+            if not isinstance(v, Path):
+                Path(v)
```
thibaultdvx:
```suggestion
                v = Path(v)
```
?

clinicadl/utils/enum.py
```diff
@@ -274,3 +274,25 @@ class RegressionMetric(str, Enum):  # TODO : put in metric module
...
+class MaskChecksum(str, Enum):
+    """Available hecksum pipelines in ClinicaDL."""
+
+    AD = "2100d514a3fabab49fe30702700085a09cdad449bdf1aa04b8f804e238e4dfc2"
+    BVFTD = "5a0ad28dff649c84761aa64f6e99da882141a56caa46675b8bf538a09fce4f81"
+    LVPPA = "1099f5051c79d5b4fdae25226d97b0e92f958006f6545f498d4b600f3f8a422e"
+    NFVPPA = "9512a4d4dc0003003c4c7526bf2d0ddbee65f1c79357f5819898453ef7271033"
+    PCA = "ace36356b57f4db73e17c421a7cfd7ae056a1b258b8126534cf65d8d0be9527a"
+    SVPPA = "44f2e00bf2d2d09b532cb53e3ba61d6087b4114768cc8ae3330ea84c4b7e0e6a"
+    AAL2 = "89427970921674792481bffd2de095c8fbf49509d615e7e09e4bc6f0e0564471"
```
thibaultdvx, line 298: Out of curiosity, what are they?

thibaultdvx (commented):
Some minor comments, LGTM otherwise!

clinicadl/utils/enum.py
camillebrianceau: They are keys to download data on the Aramis website.

PR #588, opened 2024-05-22
Trainer adaptation to data class

camillebrianceau:
- This PR aims to adapt the Training pipelines to the config classes and so `train from_json`, `resume`, `transfer_learning`, and `random-search`.
- The goal is that each argument belongs to a data class (ex: `OptimizerConfig`) and there is only one click.option corresponding to this argument with a single default value.
- I also added a `create_parameters_dict` function with all the âtch we need for now but it will be deleted at the end of the refactoring.


clinicadl/train/resume/resume.py
```diff
@@ -26,8 +38,14 @@ def automatic_resume(model_path: Path, user_split_list=None, verbose=0):
...
         config_dict["track_exp"] = None
-    if not config_dict["label_code"]:
+    if "label_code" not in config_dict or config_dict["label_code"] is None:
         config_dict["label_code"] = {}
+    if "preprocessing_json" not in config_dict:
+        config_dict["preprocessing_json"] = config_dict["preprocessing_dict"][
+            "extract_json"
+        ]
+    config_dict["maps_dir"] = model_path
+    print(config_dict)
```
thibaultdvx:
```suggestion
```

clinicadl/utils/enum.py
```diff
@@ -234,9 +234,16 @@ class ReconstructionLoss(str, Enum):  # TODO : put in loss module
 class Normalization(str, Enum):  # TODO : put in model module
     """Available normalization layers in ClinicaDL."""
 
-    BATCH = "batch"
+    BATCH = "BatchNorm"
     GROUP = "group"
-    INSTANCE = "instance"
+    INSTANCE = "InstanceNorm"
```
thibaultdvx, line 239: It seems that there are two different Normalization enum objects.

https://github.com/aramis-lab/clinicadl/blob/bb94468508087741f27f64ec94038087d7d8c7d1/clinicadl/train/tasks/reconstruction/reconstruction_config.py#L28-L33

and 

https://github.com/aramis-lab/clinicadl/blob/bb94468508087741f27f64ec94038087d7d8c7d1/clinicadl/random_search/random_search_config.py#L17-L23

I'm pretty confident we can merge them. Let's do this in the following PR. WDYT @camillebrianceau?

clinicadl/train/resume/resume_cli.py
```diff
@@ -1,22 +1,19 @@
...
+@cross_validation.split
+def cli(**kwargs):
     """Resume training job in specified maps.
 
     INPUT_MAPS_DIRECTORY is the path to the MAPS folder where training job has started.
     """
     from .resume import automatic_resume
 
-    automatic_resume(input_maps_directory, user_split_list=split)
+    automatic_resume(kwargs["input_maps_directory"], user_split_list=kwargs["split"])
```
thibaultdvx: Not sure that it is simpler to use kwargs when there are only two arguments. WDYT?

thibaultdvx (commented):
LGTM, thanks @camillebrianceau. There is just a `print` you missed. I have also noticed that there is two different enum objects for normalization layers. I think we should quickly make a PR to fix this.

clinicadl/utils/enum.py
camillebrianceau: Yes perfect

PR #590, opened 2024-05-23
[CI] Test train workflow on GPU machine

NicolasGensollen:
Try to execute the "train" tests on the GPU machine Garonne through GitHub Actions rather than Jenkins.

Related PRs: #573 and #581

### DIFF
--- a/.github/workflows/test_train.yml
+++ b/.github/workflows/test_train.yml
@@ -2,9 +2,9 @@ name: Train Tests (GPU)
 
 on:
   push:
-    branches: ["dev"]
+    branches: ["dev", "refactoring"]
   pull_request:
-    branches: ["dev"]
+    branches: ["dev", "refactoring"]
 
 permissions:
   contents: read
--- a/.github/workflows/test_train.yml
+++ b/.github/workflows/test_train.yml
@@ -19,6 +19,7 @@ env:
 
 jobs:
   test-train-gpu:
+    if: github.event.pull_request.draft == false
     runs-on:
       - self-hosted
       - Linux
--- a/.github/workflows/test_train.yml
+++ b/.github/workflows/test_train.yml
@@ -44,9 +44,9 @@ jobs:
           --junitxml=./test-reports/test_train_report.xml \
           --disable-warnings \
           --verbose \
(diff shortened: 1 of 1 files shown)


NicolasGensollen:
Pipeline output is visible here: https://github.com/aramis-lab/clinicadl/actions/runs/9204417039/job/25317904536?pr=590

Seems to be working fine.

.github/workflows/test_train.yml
```diff
@@ -0,0 +1,52 @@
+name: Train Tests (GPU)
+
+on:
+  push:
+    branches: ["dev"]
+  pull_request:
+    branches: ["dev"]
```
thibaultdvx: Do tests will also be run for PR in other branches (e.g. our refactoring branch)?

thibaultdvx (commented):
Thanks @NicolasGensollen. LGTM!

.github/workflows/test_train.yml
NicolasGensollen: Good question. With this configuration, they will only be triggered on the `dev` branch, or on PRs targeting the `dev` branch. So, they won't run on the `refactoring` branch, and they won't run on PRs targeting the `refactoring` branch.
If this is something that we want, I can make the required changes.

thibaultdvx: This is what is currently done and I think this is helpful for our refactoring. 

And I might be asking too much, but is there a way to prevent CI tests from running for a draft PR? Because we are currently making a lot of PRs and I feel that the CI is often saturated. Therefore, I would suggest that we run tests only when the PR is ready (and it is up to the developer to run tests locally when the PR is in draft mode).
WDYT @NicolasGensollen and @camillebrianceau?

NicolasGensollen: Alright, I'll make the required changes to run the CI on the `refactoring` branch as well as on PRs targeting it.

On your second point, I totally agree with you and I think this is a great suggestion. 👍 
I'm going to try a few things on PR #592 to make it work as we'd want.

thibaultdvx: Alright, sounds good. Thanks! 👍

thibaultdvx (approved):
LGTM

PR #596, opened 2024-05-28
Add flags to run CI tests locally

thibaultdvx:
I propose to add two flags to the pytest command lines to enable developers to run functional tests locally without changing the CI data:
- `--no-gpu` enables to run successfully a test on a cpu-only device without changing `gpu` to `false` in the `maps.json` file.
- `--adapt-base-dir` enables to run successfully a test without changing the paths in the `maps.json` file.

A developer could then run a test with something like:
```bash
poetry run pytest --verbose \ 
 --basetemp=/tmp/train \
 --input_data_directory=/Users/thibault.devarax/Desktop/code/clinicadl_data_ci/data_ci/ \
 --no-gpu \
 --adapt-base-dir \
 tests/test_train_ae.py
```

These two options work by virtually modifying the CI data (e.g. `gpu` will be set to `false`). So, I do agree that this is risky. But, I find this solution better than concretely modifying the CI data.
Let me stress that these flags should not be used for the final CI tests run on GitHub and should be only helpful when a developer wants to run the tests locally.

Anyways, this is just a suggestion, so feel free to comment or offer other solutions. Once we agree on a solution, I will extend it to all the tests.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -15,20 +15,21 @@ def pytest_addoption(parser):
         help="Directory for (only-read) inputs for tests",
     )
     parser.addoption(
-        "--simulate-gpu",
+        "--no-gpu",
         action="store_true",
-        help="""To simulate the presence of a gpu on a cpu-only device. Default is False.
+        help="""To run tests on cpu. Default is False.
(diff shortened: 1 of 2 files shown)


camillebrianceau (approved):
LGTM!

PR #604, opened 2024-05-30
[CI] Allow skipping GPU test pipelines with a tag in commit message

NicolasGensollen:
The goal is to enable the user to skip all GPU test pipelines manually by adding "[skip gpu]" in the commit message.


thibaultdvx:
Implemented in #933

PR #610, opened 2024-06-03
Fix issue in compare_folders

thibaultdvx:
1.  `compare_folders` used to not raise an error when it compared two empty directories. So, I added a check to make sure that the two directories to compare do exist. Consequently, I had to solve issues that appear in `test_train_ae` and `test_predict` with non existing folders.
2.  I also made changes in `test_predict.py` and `test_random_search.py` because the flags introduced in #608 didn't work well with these tests.
3.  I also added indent in the json files stored in the temporary folders created by the tests.

This PR needs modifications in CI data: because of the bug spotted in 1., the tests of the `test_train_ae` pipeline used to pass, while some reference data are missing.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -14,10 +14,19 @@ def pytest_addoption(parser):
         action="store",
         help="Directory for (only-read) inputs for tests",
     )
+    parser.addoption(
+        "--simulate-gpu",
+        action="store_true",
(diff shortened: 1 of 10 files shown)


thibaultdvx:
The tests were run with data in aramis-lab/clinicadl_data_ci#38

camillebrianceau (approved):
LGTM !

PR #612, opened 2024-06-04
Put resume in trainer

thibaultdvx:
- Creates `from_json` and `from_maps` builders in `Trainer`, in order to create an instance from a `json` file or a `MAPS` folder.
- Puts `resume` function in the `Trainer`.
- Changes `resume` and `from_json` CLI to use these new functionalities.


camillebrianceau (commented):
LGTM !

PR #616, opened 2024-06-07
Clean tasks cli and add update_with_toml method in TrainConfig

thibaultdvx:
- Cleans Classification, Regression and Reconstruction CLIs
- Adds a `update_from_toml` method in TrainConfig so that the user can update its config class with a toml config file.
- Adds a unittest function to test `update_from_toml`
- Makes Normalization options uniform accross ClinicaDL (formerly, there were "batch" and "BacthNorm"  for example)
- Updates unittets (many of them were not passing)
- Simplifies `get_type_from_config_class`, which was unnecessarily complicated


camillebrianceau (approved):
LGTM!

PR #620, opened 2024-06-07
Run unittets and cleaning

thibaultdvx:
- modify `.yaml` files to run unittests on `refactoring` (see #618)
- modify `.yaml` files to add `ready_for_review` event to trigger tests (see #623)
- removes unnecessary imports


.github/workflows/lint.yaml
camillebrianceau, line 1: Maybe we can add `mood24` branch, WDYT?

camillebrianceau (approved):
LGTM !

.github/workflows/lint.yaml
thibaultdvx: Good idea, I will make the change directly on `mood24`!

PR #623, opened 2024-06-12
Trigger tests when undrafted

thibaultdvx:
Currently, functional tests are not triggered when a PR is in draft mode. However, mark it as ready for review doesn't trigger the tests and it is necessary to push a fake commit to launch them.

I propose to specify the events that will trigger tests in a PR, and especially add the `ready_for_review` event. Formerly, the events were the [default ones](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows), i.e. `opened`, `synchronize` and `reopened`.

To summarize, if this PR is accepted:
- unittests and linter tests would always be run (even in draft mode),
- functional tests would not run in draft mode, but mark a PR as ready for review will trigger them.

### DIFF
--- a/.github/workflows/test_cli.yml
+++ b/.github/workflows/test_cli.yml
@@ -5,6 +5,7 @@ on:
     branches: ["dev", "refactoring"]
   pull_request:
     branches: ["dev", "refactoring"]
+    types: [opened, edited, reopened, ready_for_review]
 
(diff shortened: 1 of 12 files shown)


NicolasGensollen (approved):
Good idea @thibaultdvx ! 👍 
LGTM !

PR #629, opened 2024-06-24
Some cleaning

camillebrianceau:
- remove `check_gpu` from `cmdline_utils.py` and deletye the file
- adapt qc to config


thibaultdvx (approved):
LGTM

PR #633, opened 2024-06-25
Some cleaning with utils files

camillebrianceau:


clinicadl/commandline/modules_options/computational.py
```diff
@@ -19,8 +19,8 @@
     "this flag is already set to FSDP to that the zero flag is never actually removed.",
 )
 gpu = click.option(
-    "--no-gpu",
-    is_flag=True,
+    "--gpu/--no-gpu",
+    default=get_default("gpu", ComputationalConfig),
```
thibaultdvx, line 23: Seems to me that you made the opposite change in the previous PR (#629). Is it ok or is it a problem of rebasing?

tests/test_train_cnn.py
```diff
@@ -13,10 +13,10 @@
...
     params=[
         "slice_cnn",
-        "image_cnn",
-        "patch_cnn",
-        "patch_multi_cnn",
-        "roi_cnn",
+        # "image_cnn",
+        # "patch_cnn",
+        # "patch_multi_cnn",
+        # "roi_cnn",
```
thibaultdvx:
```suggestion
        "image_cnn",
        "patch_cnn",
        "patch_multi_cnn",
        "roi_cnn",
```

Is it an oversight?

thibaultdvx (commented):
Just left two comments, LGTM otherwise :)

clinicadl/commandline/modules_options/computational.py
camillebrianceau: Yes it was a mistake to do the opposite change !

PR #639, opened 2024-07-22
Clinicadl models

thibaultdvx:
This PR aims to initiate the refactoring of the network module. In this PR, I gathered all the code that concern PyTorch's `nn.Module` and reorganize it in `layers`, `blocks` and `networks`. So, there is nothing new in this PR, only some cleaning. I also added unittests to test all those architectures.

As decided during the last developper meeting, those architectures will be removed, so I didn't find it relevant to link them to the rest of `clinicadl`. That's why the folder `nn` is completely independent of the rest of the software, ready to be removed.


camillebrianceau (approved):
LGTM !

PR #640, opened 2024-07-22
Loss functions module

thibaultdvx:
In this PR, I suggest a new module for loss functions. The module is very simple and is actually only an intermediary between `ClinicaDL` and PyTorch's `nn` module. More precisely, there is:
- a `LossConfig` class, with all the parameters useful for PyTorch's loss functions,
- a factory function `get_loss_function` that will get a PyTorch loss function and parametrize it thanks to a `LossConfig` object. This function also returns a config dict with only the parameters relevant to the wanted loss function (whereas a `LossFunction` object contains the parameters for all loss functions available), because one only wants to save the information related to the wanted loss function in the MAPS.


thibaultdvx:
Update: I added an option "DefaultFromLibrary" in the config class to use by default the default values from PyTorch's loss functions, instead of manually copy/paste these defaults in our config class.

clinicadl/losses/factory.py
```diff
@@ -0,0 +1,50 @@
...
+    if "weight" in config_dict and config_dict["weight"] is not None:
+        config_dict_["weight"] = torch.Tensor(config_dict_["weight"])
+    loss = loss_class(**config_dict_)
+
+    config_dict["loss"] = config.loss
+
+    return loss, config_dict
+
+
+# TODO : what about them?
```
camillebrianceau: ??

clinicadl/losses/factory.py
```diff
@@ -0,0 +1,50 @@
...
+    expected_args, config_dict = get_args_and_defaults(loss_class.__init__)
+    for arg, value in config.model_dump().items():
+        if arg in expected_args and value != DefaultFromLibrary.YES:
+            config_dict[arg] = value
+
+    config_dict_ = deepcopy(config_dict)
+    if "weight" in config_dict and config_dict["weight"] is not None:
+        config_dict_["weight"] = torch.Tensor(config_dict_["weight"])
+    loss = loss_class(**config_dict_)
+
```
camillebrianceau, line 41: As the loss is already initialized here, do we need to return a dictionary? Because I thought we wanted to get rid of all the dictionaries.

camillebrianceau (commented):
Thanks, LGTM !

clinicadl/losses/factory.py
thibaultdvx: Hi @camillebrianceau, thanks for the review. You're right, I changed that to put the arguments of the loss function in a config class, rather than a dict. 

Even if the loss is initialized, we need the config class to store the config in the MAPS. 

The difference between the input and the output config class is that the values set to "DefaultFromLibrary" have been changed with their effective values. It is not essential but better for reproducibility I think (for example in the - unlikely - event  that PyTorch changes some default values).

Tell me what do you think!

clinicadl/losses/factory.py
thibaultdvx: Are you talking about the other losses?

I've just added more losses from PyTorch in my last commit.

PR #643, opened 2024-08-02
Optim module

thibaultdvx:
Creates the module `optim` with:
- an **optimizer** module, on the same model that the loss function module (#640): `OptimizerConfig` checks the arguments and the `get_optimizer` function gets the proper optimizer from PyTorch. 
The module may seem a bit complicated because I wanted to enable the user to pass different parameters for different parameter groups (e.g. `lr={'params_0': 0.01, 'params_1': 0.001}`), which requires more steps in the factory. Remark: the new `OptimizerConfig` is an enhanced version of the old one.
- a **LR scheduler** module, on the same model, to get the schedulers from PyTorch. Remark: the new `LRSchedulerConfig` is an enhanced version of the old one.
- an **early stopping** module: I took the old `ClinicaDL` early stopper and adds the corresponding config class and unittests to it.
- an `OptimizationConfig` class that gathers the parameters related to the optimization process (early stopping, number of epochs, etc.). Remark: this class is a merge between the old `EarlyStoppingConfig` and `OptimizationConfig`.

I'm waiting for this PR to be merged to get rid of the current `optimizer` module.


camillebrianceau (commented):
LGTM  !

PR #647, opened 2024-08-29
Monai networks

thibaultdvx:
- adds `monai` as a dependency (in `poetry.lock` and `poetry.toml`), as well as the extras `einops` and `huggingface_hub` which are required for some networks.
- creates a config class for every network we want to use from MONAI (it was too complicated to create a unique config class as was done for the optimizers or the loss functions (see #643 and #640)). The following MONAI networks would be supported with this PR (see https://docs.monai.io/en/stable/networks.html#nets): 

    - Regressor
    - Classifier
    - Critic
    - Discriminator
    - AutoEncoder
    - VarAutoEncoder
    - DenseNet
    - FullyConnectedNet
    - VarFullyConnectedNet
    - Generator
    - ResNet
    - ResNetFeatures
    - SegResNet
    - UNet
    - AttentionUnet
    - ViT
    - ViTAutoEnc      

     They seemed to be the most "famous" networks that MONAI implements, so I think they are a good basis. Of course this 
     list can be discussed.
- creates a factory function named `create_network_config`(in `monai_networks/config/factory.py`) to get the right config class from the name of the network.
- creates a getter function named `get_network` (in `monai_networks/factory.py`) to get a MONAI neural network from a config class.
- tests all these new features with unittests


camillebrianceau (approved):
LGTM !

PR #648, opened 2024-08-30
Remove TaskManager

camillebrianceau:


clinicadl/metrics/metric_module.py
```diff
@@ -32,6 +33,7 @@ def __init__(self, metrics, n_classes=2):
             for method_name in dir(MetricModule)
             if callable(getattr(MetricModule, method_name))
         ]
+        print(metrics)
```
thibaultdvx: Is it an oversight or is it on purpose?

thibaultdvx (commented):
LGTM, except a print!

clinicadl/metrics/metric_module.py
thibaultdvx:
```suggestion
```

PR #649, opened 2024-09-03
Add the word "adn" in codespell ignore words list 

camillebrianceau:
* Add the word "adn" in codespell ignore words list 
* run poetry update


thibaultdvx (approved):
LGTM, thanks!

PR #653, opened 2024-09-12
remove non in implemented reductions

thibaultdvx:
It is not possible to call `.backward()` on non-scalar.


camillebrianceau (approved):
LGTM !

PR #654, opened 2024-09-19
New metrics module based on MONAI

thibaultdvx:
This PR creates a module to get the metrics implemented in MONAI, on the same model than #647. The following metrics would be supported (see https://docs.monai.io/en/stable/metrics.html):

- Classification: 

  - Confusion Matrix related (precision, f1 score, etc.)
  - ROC AUC

- Regression
  - MAE
  - MSE
  - RMSE
  - PSNR

- Reconstruction
  - SSIM
  - Multi-scale SSIM
  - regression metrics above

- Segmentation
  - Dice
  - Generalized Dice
  - IoU
  - Surface distance
  - Hausdorff distance
  - Surface dice

- Generation
  - MMD

I'm waiting for `R2 score` and `Average Precision` to be implemented in MONAI.

---

Waiting for previous PRs to be merged to rebase.

PR #655, opened 2024-09-19
Review of loss module

thibaultdvx:
This PR is a review of the loss function module created in #640. 

More precisely, I created one config class for each supported loss function, whereas before there was a single config class for all the losses. This enables to create a `create_loss_config` function that is similar to `create_network_config` (see #647) and `create_metric_config` (see #654), and therefore have a consistent behavior across all modules.

I also put all the `Enum` class in a specific file.


camillebrianceau (approved):
LGTM !

PR #656, opened 2024-09-20
Review optim module

thibaultdvx:
This PR is a review of the optim module created in #643 . 

More precisely, I created one config class for each supported optimizer and LR scheduler, whereas before there was a single config class for all the optimizers and one for all the schedulers. This enables to create `create_optimizer_config` and `create_lr_scheduler_config` functions that are similar to `create_network_config` (see #647), `create_metric_config` (see #654) and `create_loss_config` (see #655) and therefore have a consistent behavior across all modules.

I also put all the `Enum` class in specific files.


camillebrianceau (approved):
LGTM !

PR #657, opened 2024-09-26
Sortir les étapes de validation du MapsManager

camillebrianceau:
In this PR, I aim to remove the validation methods from the Maps Manager:

- I created a Validator class and moved the validation methods there.
- I created a ValidatorConfig, although it is not used yet.
- I added `API_test.py`, which serves as a model for the desired structure of the API.

Other change for the Validator will be in another PR


clinicadl/validator/validator.py
```diff
@@ -0,0 +1,512 @@
...
+        # split_name,
+        # num_networks,
+        # mode,
+    ):
+        # maps_manager.maps_path = maps_path
+        # maps_manager.split_name = split_name
+        # maps_manager.num_networks = num_networks
+        # maps_manager.mode = mode
+
+        print("test")
```
thibaultdvx: Is it useful?

thibaultdvx (commented):
LGTM, thanks!

PR #658, opened 2024-09-27
Clean MapsManager 

camillebrianceau:
This PR refactors the training methods in the `trainer` module.

- First, I removed the `_train_multi` method. Now, the `train_single` method is used for both single and multi-network training. In a future PR, we will need to apply the same approach to the `_train_ssda` method.
- I also moved the loops `for split in split_manager.iterate():` and `for network in range(num_network):` out of the `train_single` method. As a result, the new `_train` method will be the entry point for `trainer.train()`.
- Additionally, I added an example of the future API in `API_test.py`.
- I created a `TmpConfig`, which is currently unused but retained as a model for future reference.

### Next PR:
- Plan to extract the `MapsManager`, `Validator`, and `Splitter` from the `trainer`.


clinicadl/trainer/tasks_utils.py
```diff
@@ -580,6 +580,10 @@ def check_prediction(row):
         ["participant_id", "session_id"]
     ):
         label = subject_df["true_label"].unique().item()
+        print(subject_df)
+        for i in range(n_classes):
+            print(subject_df[f"proba{i}"])
+        print(weight_series)
```
thibaultdvx: On purpose?

thibaultdvx (commented):
LGTM, except suspect prints.

PR #660, opened 2024-10-01
Basic neural networks

thibaultdvx:
I this PR, I suggest the basic neural network architectures that could be trained via the commande line. @camillebrianceau @ravih18 @msolal @HuguesRoy could you please have a look at it?

The implemented networks are the following:
- `MLP`: a Multi Layer Perceptron (or Fully Connected Network), where linear, activation, normalization and dropout layers can be customized.
- `ConvEncoder`: a Fully Convolutional Encoder, where convolutional, pooling, activation, normalization and dropout layers can be customized.
- `ConvDecoder`: a Fully Convolutional Decoder, very similar to `ConvEncoder` but convolutions are replaced by transposed convolutions and pooling layers by unpooling layers.
- `CNN`: a regressor/classifier with first a convolutional part  and then a fully connected part. It is a simple aggregation of a `ConvEncoder` and a `MLP`, so it is entirely customizable.
- `Generator`: the symmetrical network of `CNN`, made with the aggregation of a `MLP` and a `ConvDecoder`.
- `AutoEncoder`: a symmetrical autoencoder, built from the aggregation of a `CNN` (encoder) and the corresponding `Generator` (decoder).
- `VAE`: a Variational AutoEncoder, built on the same idea as `AutoEncoder`.

Please look at the docstrings for some examples.
All the architectures are tested with unittests.

Feel free to comment the choices I made. To me, the goal is to have sufficiently flexible architectures to enable the user to tune their parameters, without going into too complicated neural networks that could be implemented via the API.
Feel also free to comment my choices on network names.


thibaultdvx:
Since last comment, I have added well-know networks from literature:
- [ResNet](https://arxiv.org/pdf/1512.03385)
- [DenseNet](https://arxiv.org/pdf/1608.06993)
- [Squeeze-and-Excitation ResNet](https://arxiv.org/pdf/1709.01507)
- [Vision Transformer](https://arxiv.org/pdf/2010.11929)
- [UNet](https://arxiv.org/pdf/1505.04597)
- [Attention UNet](https://arxiv.org/pdf/1804.03999)

These networks are also customizable. For example, in `DenseNet`, the user can specify the number of dense blocks, as well as the number of dense layers in each block. 
However, if the user doesn't want to customize the network and wants to have access to the networks with the hyper-parameters used in the reference paper, the user can use factory functions, e.g. `get_densenet("DenseNet-121")`. Furthermore, if pre-trained weights are available via `torchvision`, the user can download them.

The implementations mostly rely on `MONAI` and `torchvision`. The are all tested with unittests. Please look at the docstrings for some examples.

thibaultdvx:
@camillebrianceau @ravih18 
All right guys, I'm quite happy with this version of the network module. Let me know if you have any comment.
Otherwise, I will create the config classes associated to the new networks.

thibaultdvx:
Last commits, I changed the config classes to match the new networks. I also changed the factory function `get_network`.
Now it can be used in the following way: `get_network(name="DenseNet", spatial_dims=1, n_dense_layers=(6, 12, 24))`, with first the name of the network and then kwargs to parametrize the network.

camillebrianceau (approved):
LGTM !

PR #666, opened 2024-10-09
Cb extract validator

camillebrianceau:
In this PR:
- The `caps_directory` is now optional because it is not needed for the `predict` and `interpret` commands when the data group is already defined.
- Merged `PredictManager` and `Validator` into a single class, temporarily named `Predictor`.
- Removed `predict_multi` as multi-network prediction can now be handled in `predict_single`.
- Renamed `SplitManager` to `Splitter`.
- Initialized the predictor and splitter in `trainer.__init__()`.
- Removed SSDA


thibaultdvx (approved):
LGTM!

PR #679, opened 2024-11-06
FileType change in clinicadl_file_reader (minor change)

camillebrianceau:
Ready for review

Update the `clinica_utils` file so that it uses the `FileType` object instead of the `information` dictionary


clinicadl/transforms/transforms.py
```diff
@@ -6,9 +6,9 @@
 class Transforms:
     def __init__(
         self,
-        data_augmentation=List[torchio],
-        image_transforms=List[torchio],
-        object_transforms=List[torchio],
+        data_augmentation=List[torchio.Transform],
+        image_transforms=List[torchio.Transform],
+        object_transforms=List[torchio.Transform],
```
thibaultdvx:
```suggestion
        data_augmentation=List[Callable],
        image_transforms=List[Callable],
        object_transforms=List[Callable],
```

The user should be able to user other transforms than torchio's.

thibaultdvx (commented):
LGTM

PR #683, opened 2024-11-12
Uniformize modules

thibaultdvx:
To uniformize `losses`, `optimizers`, `lr_schedulers`, `metrics` and `networks` modules. They all end up with two main functions like: 
- `get_network_config` (e.g. `network_config=get_network_config("MLP", num_inputs=2)`), so that the user can pass his parameters in a config object.
- `get_network_from_config` (e.g. `network=get_network_from_config(network_config)`), so that we can easily retreive the right object from a config.

All 5 modules are tested with 100% coverage.


camillebrianceau (approved):
LGTM !

PR #684, opened 2024-11-26
First draft for KFold

thibaultdvx:
Here is a first draft for `KFold`.

At initialization, `KFold` takes one or several `CapsDataset` (e.g. `KFold(dataset)` or `KFold(dataset_0, dataset_1)`).
The public methods are:
- `make_splits`: to perform the splitting operation. This method should do the job of `clinicadl kfold`;
- `read`: to get the splits from a split directory;
- `write`: after `make_splits` has been called, to write the splits in a split directory;
- `get_splits`: returns an iterator of the wanted splits (e.g. `kfold.get_splits(splits=[0, 3])`).

The splits are returned as a `Split` object that contains all the relevant information for training, including the training and validation datasets.  If multiple datasets have been passed at the initialization of the `KFold` object, a tuple of `Split` objects is returned (splits are arranged according to the order of input datasets).

To build the associated training and validation dataloaders, the user must call the `build_train_loader` and `build_val_loader` methods of the `Split` objects. These methods accept common dataloader parameters (`batch_size`, `shuffle`, etc.). I chose to separate these two methods because the user may want different parameters for his/her training and validation loaders.

I also felt the need to refactor the `get_dataloader` and `generate_sampler` functions.

Beware that it is a draft and nothing is yet tested.


camillebrianceau:
As we discussed:

`make_kfold` or `make_single_split`: to perform the splitting operation (takes a tsv file to split)

At initialization, KFold / SingleSplit takes a `split_directory` with a json file inside resuming the split.

The only public method is finally `get_splits` which returns an iterator of the wanted splits (e.g. `kfold.get_splits(dataset = CapsDataset(), splits=[0, 3])`).
The splits are returned as a Split object containing all the relevant training information, including the training and validation datasets. 
For multiple datasets, we will use `ConcatDataset` or `StackDataset`

I haven't touched the DataLoader.

clinicadl/experiment_manager/experiment_manager.py
```diff
@@ -8,15 +8,14 @@
...
-from clinicadl.dataset.preprocessing import PreprocessingConfig
+from clinicadl.dataset.preprocessing import BasePreprocessing
+from clinicadl.dataset.readers import CapsReader
 from clinicadl.metrics.old_metrics.utils import check_selection_metric
 from clinicadl.model.clinicadl_model import ClinicaDLModel
 from clinicadl.networks.config import NetworkConfig
 from clinicadl.networks.factory import get_network_from_config
-from clinicadl.splitter.kfold import KFolder
 from clinicadl.splitter.split_utils import print_description_log
+from clinicadl.transforms.extraction import BaseExtraction
```
thibaultdvx:
```suggestion
from clinicadl.transforms.extraction import Extraction
```

clinicadl/dataset/datasets/caps_dataset.py
```diff
@@ -535,3 +543,28 @@ def prepare_image(participant, session):
...
         )
+
+    def subset(self, data: Optional[Union[pd.DataFrame, Path]] = None) -> CapsDataset:
+        df = self._check_data_instance(data)
+
+        common_rows = pd.merge(df, self.df, how="inner")
+        all_included = len(common_rows) == len(df)
+
+        if not all_included:
+            missing_rows = pd.concat([df, self.df], ignore_index=True).drop_duplicates(
```
thibaultdvx: I may be wrong but it seems to me that you will have also the rows that are in `self.df` but not in `df`, whereas you only want rows that are in `df` but not in `self.df`, don't you ?

clinicadl/dataset/datasets/caps_dataset.py
```diff
@@ -535,3 +543,28 @@ def prepare_image(participant, session):
...
+            err_message = "Missing rows: \n"
+            for row in missing_rows:
+                err_message += f" - {row} \n"
+
+            raise ClinicaDLTSVError(
+                "Some couples (participanst_id, session_id) are not in the dataset,",
+                err_message,
+            )
+
+        dataset = copy(self)
```
thibaultdvx: I'm wondering if a deepcopy is not better here. WDYT?

clinicadl/splitter/make_splits/kfold.py
```diff
@@ -0,0 +1,185 @@
...
+            raise ValueError(
+                "Continuous variables cannot be used for stratification in K-Fold splitting."
+            )
+
+    return df[[column]]
+
+
+def make_kfold(
+    tsv_path: Path,
+    output_dir: Optional[Path] = None,
```
thibaultdvx: As this function will be in our API, I think it would be nice if it accepts also strings.

clinicadl/splitter/make_splits/kfold.py
```diff
@@ -0,0 +1,185 @@
...
+        Path to the input TSV file.
+    output_dir : Optional[Path]
+        Directory to save the split files. Defaults to the parent directory of `tsv_path`.
+    subset_name : str, default="validation"
+        Name of the subset used for output files.
+    valid_longitudinal : bool, default=False
+        Whether to include longitudinal sessions in the split.
+    n_splits : PositiveInt, default=5
+        Number of splits for K-Fold.
+    stratification : Optional[List[str]], default=None
```
thibaultdvx: A list or a single str ?

clinicadl/splitter/make_splits/kfold.py
```diff
@@ -0,0 +1,185 @@
...
+        if column not in df.columns:
+            raise ClinicaDLConfigurationError(
+                f"Stratification column '{column}' not found in the dataset."
+            )
+        return column
+    if not stratification and not ignore_demographics:
+        raise ClinicaDLConfigurationError(
+            "No stratification column specified. Provide a column or enable `ignore_demographics`."
+        )
+    return None
```
thibaultdvx: There is something I'm missing: what is the role of `ignore_demographics`?
Is it not enough to pass `stratification=None` when one doesn't want stratification?

clinicadl/splitter/make_splits/kfold.py
```diff
@@ -0,0 +1,185 @@
...
+
+
+def make_kfold(
+    tsv_path: Path,
+    output_dir: Optional[Path] = None,
+    subset_name: str = "validation",
+    valid_longitudinal: bool = False,
+    n_splits: PositiveInt = 5,
+    stratification: Optional[str] = None,
+    ignore_demographics: bool = False,
```
camillebrianceau:
```suggestion
    ignore_demographics: bool = True,
```

clinicadl/splitter/make_splits/single_split.py
```diff
@@ -0,0 +1,440 @@
...
+        p_categorical_threshold=p_categorical_threshold,
+        stratification=stratification,
+        ignore_demographics=ignore_demographics,
+    )
+
+    config._check_split_dir()
+    config._write_json()
+
+    if config.n_test > 0:
+        splits = ShuffleSplit(n_splits=n_try_max, test_size=config.n_test)
```
thibaultdvx: Here should we add a `random_state` like in `make_kfold`?

clinicadl/splitter/make_splits/single_split.py
```diff
@@ -0,0 +1,440 @@
...
+        for n_try, (test_index, train_index) in enumerate(
```
thibaultdvx: `(train_index, test_index)`, no?

clinicadl/splitter/make_splits/single_split.py
```diff
@@ -0,0 +1,440 @@
...
+                        baseline_df,
+                        subset_name,
+                    )
+                    break
+
+        if n_try >= n_try_max - 1:
+            raise ClinicaDLConfigurationError(
+                f"Unable to find a valid split after {n_try} attempts. "
+                f"Consider lowering thresholds or reducing stratification variables."
+            )
```
thibaultdvx: This is not of the utmost importance but I think there is an indent issue here. Because if the last split is the good one, we get out of the loop with `n_try=n_try_max - 1` and it will raise the issue.
Whereas if we keep that block inside the for loop, after `break`, it should be ok.

clinicadl/splitter/make_splits/kfold.py
```diff
@@ -0,0 +1,185 @@
...
+    column = _validate_stratification(df, stratification)
+
+    if column is None:
+        return df
+
+    if pd.api.types.is_numeric_dtype(df[column]):
+        if len(np.unique(df[column])) >= n_test:
+            raise ValueError(
+                "Continuous variables cannot be used for stratification in K-Fold splitting."
+            )
```
thibaultdvx: Already done in `_validate_stratification`, no?

clinicadl/splitter/make_splits/kfold.py
```diff
@@ -0,0 +1,185 @@
...
+
+    raise ClinicaDLConfigurationError(
+        "Invalid or conflicting stratification options provided. Stratification must be a single column name or boolean."
+    )
+
+
+def preprocess_stratification(
+    df: pd.DataFrame,
+    stratification: Union[str, bool],
+    n_test: int = 100,
```
thibaultdvx: This `n_test` seems weird to me. It is always 100, no matter the length of `df`.

clinicadl/splitter/splitter/splitter.py
```diff
@@ -0,0 +1,262 @@
...
+        """
+        Initialize Split with a dataset.
+
+        Parameters
+        ----------
+        dataset : CapsDataset
+            Dataset to split for cross-validation.
+        """
+        if isinstance(split_dir, str):
+            split_dir = Path(split_dir)
```
thibaultdvx:
```suggestion
        split_dir = Path(split_dir)
```
?

clinicadl/splitter/splitter/splitter.py
```diff
@@ -0,0 +1,262 @@
...
+            Directory containing the JSON configuration file.
+
+        Returns
+        -------
+        KFoldConfig
+            The configuration object loaded from the JSON file.
+        """
+
+        json_file = [json for json in self.split_dir.glob("*.json")]
+        print(json_file)
```
thibaultdvx:
```suggestion
```

clinicadl/splitter/splitter/single_split.py
```diff
@@ -0,0 +1,95 @@
...
+)
+
+
+class SingleSplitConfig(SplitterConfig):
+    json_name: str = "single_split_config.json"
+    subset_name: str = "test"
+    stratification: Union[List[str], bool] = False
+    n_test: PositiveInt = 100
+    p_categorical_threshold: float = 0.80
+    p_continuous_threshold: float = 0.80
```
thibaultdvx, line 21: In the future, I think we should synchronise these default values with those in `make_split`. WDYT?

thibaultdvx (commented):
Hi @camillebrianceau, thanks for the PR! 
I left some comments, reviewed my previous work on the dataloaders, turned pydantic `BaseModel` to `ClinicaDLConfig`, and add some minor changes suggested `pylint`. Tell me if you don't agree with one the previous.

Looks good to me overall!

thibaultdvx:
Last suggested change: put the part concerning the dataloader in the module `dataset`, as it is not directly linked to splits.
I'm not completely sure about this. To be discussed!

camillebrianceau:
Perfect for the `DataLoader`. I also think we should rename the `dataset` module to something more general, like `data`, or something like that.

I agree that we should synchronize the default values, but I think it’s something we can do after refactoring the remaining modules (e.g., `Trainer`,` ClinicaDLModel`, etc.).

I’ve made the change you proposed. Feel free to merge if everything looks good to you!

thibaultdvx:
Perfect, I do agree for `data`. Let's do that in a next PR!

PR #685, opened 2024-11-28
Dataset module for v2

camillebrianceau:
In this PR, I refactored the `dataset` module with the following changes:

- I kept the `API/dataset_test.py` in the PR to provide an example of how the dataset is used.
- There is now a single `CapsDataset` for all extraction modes.  
  Future implementations in ClinicaDL will introduce:  
  - `ConcatDataset`  
  - `PairedDataset`  
  - `UnPairedDataset`  
  These will inherit from `CapsDataset`.
- Preprocessing is now represented as an object (using `pydantic`). This object describes:  
  - The preprocessing steps performed using Clinica.  
  - The image type being used.
- Readers are the objects responsible for reading data from CAPS or BIDS. These will eventually be replaced by ClinicaIO once its development is complete.

The extraction process has been moved to the `Transforms` module. Each extraction type (`ROI`, `Image`, `Slice`, and `Patch`) is now represented as an object and behaves like a transform.

I added unit tests. However, tests for `Transforms` and `Preprocessing` are not included, as these components are expected to change in the coming weeks.


thibaultdvx:
> The extraction process has been moved to the `Transforms` module. Each extraction type (`ROI`, `Image`, `Slice`, and `Mask`) is now represented as an object and behaves like a transform.

You meant `Patch`, didn't you?

clinicadl/dataset/readers/multi_caps_reader.py
```diff
@@ -0,0 +1,40 @@
...
+
+import pandas as pd
+
+from clinicadl.utils.exceptions import ClinicaDLTSVError
+
+from .caps_reader import CapsReader
+from .reader import Reader
+
+
+class CapsMultiReader:
```
thibaultdvx, line 11: I'm curious, why did you feel the need to create this reader? Is it temporary?

Because I thought that we would handle multi-cohort with multiple `CapsDataset`, concatenated with `ConcatDataset`. And it seems to me that each `CapsDataset` could have its own `CapsReader`.

Tell me if I'm not clear enough!

thibaultdvx (approved):
Great work, thanks @camillebrianceau!!

clinicadl/dataset/readers/multi_caps_reader.py
camillebrianceau: Yes I didn't explain but it was just a first draft. Then we put the `CapsReader` in the `CapsDataset` so it is not useful anymore but since we’re not entirely sure about our change, I prefer not to delete it

thibaultdvx: Ok, LGTM!

PR #686, opened 2024-12-09
Continuing refactoring of the extraction objects

thibaultdvx:
Following #685, I continued the refactoring of the extraction classes `Image`, `ROI`, `Patch` and `Slice`. I suggest to get rid of some complicated features while adding some useful ones:

- For all: remove `use_uncropped_image` that should be put in `Preprocessing`.

- `Slice`: in addition to `discarded_slices` that only enable slice filtering, I suggest to introduce two new arguments. `slices`  enables slice selection (e.g. `Slice(slices=[128, 129, 130])`); and `borders` enables slice filtering at the edge of the image. For example, `Slice(discarded_slices=[42], borders=5)` will filter out the slice with index `42`, as well as the first five and the last five slices in each direction. `borders` can also be specified for each border (e.g. `Slice(borders=((1, 1), (2, 3), (1, 1))`).

- `Patch`: I suggest to enable anisotropic sliding window by allowing the user to pass anisotropic values for `patch_size` and `stride` (e.g. `Patch(patch_size=(18, 16, 17), stride=(1, 1, 2)`).

- `ROI`:  to replace `roi_list`, `roi_mask_location`, `roi_crop_input`, `roi_crop_output`, `roi_template`,  `roi_mask_pattern`, `roi_custom_template` and `roi_custom_mask_pattern`, I would suggest to have only two arguments. `masks`, where a list of masks can be passed via their file location (e.g. `ROI(masks=["masks/leftHippocampus.nii.gz", "masks/rightHippocampus.nii.gz"])`); and `crop` that enables to filter out regions that are always 0 for all masks, in order to reduce the size of the images. 
In doing so, I purposely release some constraints on the masks that can be used, that I felt unnecessary.

Besides, I also made some minor changes:
- Renaming `BaseExtraction` to `Extraction`, in order not to have `Base...` in our typings;
- Gathering `roi`, `patch`, `slice` and `image` under the generic word `sample`. I changed the function names and dosctrings accordingly;
- Uniformizing the docstring between the four child classes (there is still work the be done).

As of now, the classes are not tested. i'm waiting for your opinion first.


thibaultdvx:
The classes are now tested with 100% coverage.

thibaultdvx:
I just added some useful class methods to manipulate TorchIO's class `Subject`. This is crucial as our transforms will manipulate `Subject` objects. More precisely:
- `get_tio_image`: puts the image and the label in a `Subject`, as well as other potential masks that will be used to compute the transforms.
- `extract_tio_sample`: does the extraction like `extract_sample`, but works on `Subject`. All the images in the `Subject` object will undergo the extraction. For example, in the case of segmentation, extraction will be performed also on the  segmentation map.
- `format_output`: takes a `Subject` and some metadata and returns a `Sample`, which is typically the output of the `__getitem__` in our `CapsDataset`. `Sample` is similar to the old `CapsDatasetSample`, but it is different depending on the extraction method.

So, in the `__getitem__` of our `CapsDataset`, I imagine it this way:

```python
    participant, session, img_index, sample_index = self._get_meta_data(idx)
    image, image_path = self._get_image(img_index)
    label = self._get_label(img_index)

    tio_image = self.transforms.extraction.get_tio_image(image, label)

    tio_image = self.transforms.image_transforms(tio_image)

    if not self.eval_mode:
        tio_image = self.transforms.image_augmentation(tio_image)

    tio_sample = self.transforms.extraction.extract_tio_sample(tio_image, sample_index)

    tio_sample = self.transforms.sample_transforms(tio_sample)

    if not self.eval_mode:
        tio_sample = self.transforms.sample_augmentation(tio_sample)

    return self.transforms.extraction.format_output(
        tio_sample,
        participant_id=participant,
        session_id=session,
        image_path=image_path,
    )
```

thibaultdvx:
In `Slice`, I removed the argument `slice_mode`, as duplicating channels can easily be done with a basic transform.

thibaultdvx:
As discussed in developer meeting, I removed the ROI extraction as it can be done with common transforms (e.g. [Mask](https://torchio.readthedocs.io/transforms/preprocessing.html#mask), potentially followed by [Crop](https://torchio.readthedocs.io/transforms/preprocessing.html#crop)).

camillebrianceau (approved):
LGTM !

PR #688, opened 2024-12-17
ClinicaDL dictionary

camillebrianceau:
An idea for a new module `Dictionary` with all the str that we use repetitively in the code.
I added an example for the `Transforms` module.


thibaultdvx (approved):
Good idea, LGTM!

PR #690, opened 2024-12-18
Cleaning part 2 + DataGroup

camillebrianceau:
In this PR:

- Added a file to define custom typings (e.g., `PathType = Union[str, Path]` and `DataType = Union[pd.DataFrame, str, Path]`).  
- Created a `DataGroup` class to represent a DataGroup in the MAPS.  
- Introduced a first **DRAFT** for a `MapsReader`. This is somewhat similar to an `ExperimentManager`, but since the role of this module is not yet clear, I wanted to keep them separated.  
- Added a first **DRAFT** for a `Predictor`: nothing needs to be kept in the future. For now, it just helps clarify certain concepts.  
- Created a MAPS in the `tests/resources` directory.


clinicadl/predictor/old_predictor/old_predictor.py
```diff
@@ -1059,7 +1059,7 @@ def _compute_output_tensors(
         Compute the output tensors and saves them in the MAPS.
 
         Args:
-            dataset (clinicadl.data.datasets.caps_dataset.CapsDataset): wrapper of the data set.
+            dataset (clinicadl.dataset.caps_dataset.CapsDataset): wrapper of the data set.
```
thibaultdvx: Are you sure? `CapsDataset` are in `data/datasets/caps_dataset.py`.

thibaultdvx (approved):

clinicadl/predictor/old_predictor/old_predictor.py
camillebrianceau:
```suggestion
            dataset (clinicadl.data.datasets.CapsDataset): wrapper of the data set.
```

PR #691, opened 2024-12-23
Final transform module

thibaultdvx:
First of all, lots of outdated files deleted: `networks/old_network` and the associated tests in `tests/unittests/nn`.

### Config classes for TorchIO transforms

Supported transforms are:

- [RescaleIntensity](https://torchio.readthedocs.io/transforms/preprocessing.html#torchio.transforms.RescaleIntensity)
- [ZNormalization](https://torchio.readthedocs.io/transforms/preprocessing.html#torchio.transforms.ZNormalization)
- [Mask](https://torchio.readthedocs.io/transforms/preprocessing.html#torchio.transforms.Mask)
- [Clamp](https://torchio.readthedocs.io/transforms/preprocessing.html#torchio.transforms.Clamp)
- [CropOrPad](https://torchio.readthedocs.io/transforms/preprocessing.html#torchio.transforms.CropOrPad)
- [Resize](https://torchio.readthedocs.io/transforms/preprocessing.html#torchio.transforms.Resize)
- [Resample](https://torchio.readthedocs.io/transforms/preprocessing.html#torchio.transforms.Resample)
- [EnsureShapeMultiple](https://torchio.readthedocs.io/transforms/preprocessing.html#torchio.transforms.EnsureShapeMultiple)
- [Crop](https://torchio.readthedocs.io/transforms/preprocessing.html#torchio.transforms.Crop)
- [Pad](https://torchio.readthedocs.io/transforms/preprocessing.html#torchio.transforms.Pad)
- [RemapLabels](https://torchio.readthedocs.io/transforms/preprocessing.html#torchio.transforms.RemapLabels)
- [OneHot](https://torchio.readthedocs.io/transforms/preprocessing.html#torchio.transforms.OneHot)
- NanRemoval (from old `clinicadl`)

Supported augmentations are:

- [OneOf](https://torchio.readthedocs.io/transforms/augmentation.html#torchio.transforms.OneOf)
- [RandomFlip](https://torchio.readthedocs.io/transforms/augmentation.html#torchio.transforms.RandomFlip)
- [RandomAffine](https://torchio.readthedocs.io/transforms/augmentation.html#torchio.transforms.RandomAffine)
- [RandomElasticDeformation](https://torchio.readthedocs.io/transforms/augmentation.html#torchio.transforms.RandomElasticDeformation)
- [RandomAnisotropy](https://torchio.readthedocs.io/transforms/augmentation.html#torchio.transforms.RandomAnisotropy)
- [RandomMotion](https://torchio.readthedocs.io/transforms/augmentation.html#torchio.transforms.RandomMotion)
- [RandomGhosting](https://torchio.readthedocs.io/transforms/augmentation.html#torchio.transforms.RandomGhosting)
- [RandomSpike](https://torchio.readthedocs.io/transforms/augmentation.html#torchio.transforms.RandomSpike)
- [RandomBiasField](https://torchio.readthedocs.io/transforms/augmentation.html#torchio.transforms.RandomBiasField)
- [RandomBlur](https://torchio.readthedocs.io/transforms/augmentation.html#torchio.transforms.RandomBlur)
- [RandomNoise](https://torchio.readthedocs.io/transforms/augmentation.html#torchio.transforms.RandomNoise)
- [RandomSwap](https://torchio.readthedocs.io/transforms/augmentation.html#torchio.transforms.RandomSwap)
- [RandomGamma](https://torchio.readthedocs.io/transforms/augmentation.html#torchio.transforms.RandomGamma)

### `Transforms` object

The main object of the module, `Transforms`, now supports config classes as inputs (and still supports raw transforms as well). E.g: 

```python
>>> from clinicadl.transforms import Transforms, get_transform_config, Patch
>>> import torchio as tio
>>> transforms = Transforms(
        extraction=Patch(),
        image_transforms=[get_transform_config("ZNormalization")],
        sample_transforms=[tio.Resize((16, 16, 16))],
        image_augmentations=[get_transform_config("RandomBlur")],
        sample_augmentations=[tio.RandomAffine()],
    )
>>> transforms.get_transforms()

(Compose([ZNormalization(masking_method=None)]),
 Compose([Resize(target_shape=[16 16 16], image_interpolation=linear, label_interpolation=nearest)]),
 Compose([RandomBlur()]),
 Compose([RandomAffine()]))
```

The whole `transforms` module is tested with 100% coverage.


clinicadl/data/utils.py
```diff
@@ -259,3 +259,96 @@ def get_infos_from_parameters(
...
+            self.common_mask = True
+            self.mask = Path(mask)
+
+        elif isinstance(mask, str):
+            if self._check_path(mask):
+                self.common_mask = True
+                self.mask = Path(mask)
+            else:
+                self.common_mask = False
+                self.mask = mask
```
camillebrianceau:
```suggestion
                self.mask = Path(mask)
```

camillebrianceau (approved):
Just one thing: What do you think about doing something like this for the `Subject` class? This way, we are forced to have an `image` and a `label` in the class, while still being able to work with all the TorchIO transforms.

```
class ClinicaDLSubject(tio.Subject):
    """
    A TorchIO Subject with an 'image' attribute, a TorchIO ScalarImage and a 'label' attribute, which can be either a float, int or torch.Tensor.
    """
    image: tio.ScalarImage
    label: Union[float, int, torch.Tensor]
```
   
   Otherwise, LGTM ! Thanks :)

clinicadl/data/utils.py
thibaultdvx: In this case, `mask`  is not a path, but rather a suffix. E.g. with `Mask("brain")`, the mask will be expected in `sub-..._ses-..._brain.nii.gz`.
So, I'm not sure we should put it in a `Path` object.

thibaultdvx:
Thanks for the review @camillebrianceau. Sounds a good idea, let me check if it works!

thibaultdvx:
As you suggested @camillebrianceau, I added a subclass of TorchIO `Subject` with mandatory `image` and `label` arguments. I had to overwrite the `__copy__` method to make it works with TorchIO's transforms. I'm not a fan of the name `ClinicaDLSubject` because it is confusing with `Subject` that we are building in the `clinicaio` project. So, I suggested the name `DataPoint` (I was not really inspired haha), but please fell free to suggest if you have a better name.

thibaultdvx:
I didn't want to put `image` in the name, as it would be confusing with `ImageSample`.

thibaultdvx:
Besides, I added two TorchIO transforms I had forgotten: [Resample](https://torchio.readthedocs.io/transforms/preprocessing.html#torchio.transforms.Resample) and [ToCanonical](https://torchio.readthedocs.io/transforms/preprocessing.html#torchio.transforms.ToCanonical).

camillebrianceau:
Just a quick comment—I think we can address this later, as it seems we've been making this mistake since we started using Pydantic. The type of a parameter in a Pydantic dataclass should reflect its actual type, not all the accepted types.

For example, in the `Patch` class, `stride` and `patch_size` should be defined as:

```python
Tuple[PositiveInt, PositiveInt, PositiveInt] = (50, 50, 50)
```

Meanwhile, the field validator can accept other types, such as:

```python
Union[PositiveInt, Tuple[PositiveInt, PositiveInt, PositiveInt]]
```

I think we may have also made this mistake with `Union[Path, str]` and `Optional[something]` so this is just something to keep in mind for a future PR.

I’m not a huge fan of `DataPoint`, and since we don’t yet know when we’ll use `ClinicaIO`, I think we can use `ClinicaDLSubject` but we can do this later if needed, so let’s keep it as-is for now.

Otherwise, everything looks great, I let you merge when you want, thanks for your work on this! :)

thibaultdvx:
I think you're right and that, strictly speaking, validators should only validate and not convert to other types. I changed `Patch` and `Slice` accordingly to have concrete examples.  There may be a better way to do it, so let's move on and keep it in mind for the next PRs, as you suggested.

```python
class Patch(Extraction):

    patch_size: Tuple[PositiveInt, PositiveInt, PositiveInt]
    stride: Tuple[PositiveInt, PositiveInt, PositiveInt]

    def __init__(
        self,
        *,
        patch_size: Union[
            PositiveInt, Tuple[PositiveInt, PositiveInt, PositiveInt]
        ] = 50,
        stride: Union[PositiveInt, Tuple[PositiveInt, PositiveInt, PositiveInt]] = 50,
    ) -> None:
        super().__init__(
            patch_size=self._ensure_tuples(patch_size),
            stride=self._ensure_tuples(stride),
        )

    @staticmethod
    def _ensure_tuples(
        value: Union[PositiveInt, Tuple[PositiveInt, PositiveInt, PositiveInt]],
    ) -> Tuple[PositiveInt, PositiveInt, PositiveInt]:
        """
        Ensures that 'patch_size' and 'stride' are always tuples.
        """
        if isinstance(value, int):
            return (value, value, value)
        else:
            return value
```

PR #692, opened 2024-12-23
data module review

thibaultdvx:
Another review of `data` as there are still some issues. The goal of this PR is to have a functional and tested `data` module. 
The work of this PR relies on #691, so I will wait the rebase operation before merging.


clinicadl/data/datatype/modalities/__init__.py
```diff
@@ -1,5 +1,5 @@
 from .base import Modality
-from .custom import Custom
+from .custom import CustomModality
```
camillebrianceau: I think I just prefer 'Custom' to match other modalities (even if there is also a custom for preprocessing)

clinicadl/data/datatype/modalities/pet.py
```diff
@@ -1,47 +1,72 @@
...
+    ----------
+    tracer : Tracer
+        radioactive tracer used for acquisition.
+    reconstruction : Optional[ReconstructionMethod] (optional, default=None)
+        the method used to reconstruct the image.
     """
 
-    tracer: Tracer = Tracer.FFDG
-    suvr_reference_region: SUVRReferenceRegions = SUVRReferenceRegions.CEREBELLUMPONS2
+    tracer: Tracer
```
camillebrianceau: I think we can keep the default tracer here

clinicadl/data/datatype/preprocessing/base.py
```diff
@@ -2,15 +2,16 @@
...
 from clinicadl.utils.config import ClinicaDLConfig
 
+from ..file_type import FileType
+from ..modalities import Modality
+
 
 class Preprocessing(ClinicaDLConfig, abc.ABC):
     """
-    Abstract configuration class for the preprocessing procedure.
+    Abstract configuration class for to model the preprocessing step.
```
camillebrianceau:
```suggestion
    Abstract configuration class to model the preprocessing step.
```

clinicadl/data/datatype/preprocessing/custom.py
```diff
@@ -1,34 +1,50 @@
...
+from ..enum import PreprocessingMethod
+from ..file_type import FileType
+from ..modalities import CustomModality
 from .base import Preprocessing
 
+logger = getLogger("clinicadl.data.datatype.preprocessing.custom")
+
 
-class Custom(Preprocessing, CustomModality):
+class CustomPreprocessing(Preprocessing, CustomModality):
```
camillebrianceau: Same here, I think we can keep `Custom`. When the user uses `Modality` he doesn't use `Preprocessing` so he will not have both in the same script. WDYT ?

clinicadl/data/tensor_conversion.py
```diff
@@ -0,0 +1,885 @@
...
+
+        # are all (participant, session)s converted?
+        self._compare_participants_sessions(conversion_info)
+
+        # all checks passed, update current state
+        self._participants_sessions_converted = (
+            self.caps_dataset.get_participant_session_couples()
+        )
+        self._masks_converted = [
+            mask.path.name for mask in self.caps_dataset.common_masks
```
camillebrianceau, line 183:
```suggestion
            mask.name for mask in self.caps_dataset.common_masks
```

camillebrianceau (approved):
LGTM ! Made some changes but feel free not to accept them ! :)

clinicadl/data/datatype/preprocessing/custom.py
thibaultdvx: yes that's true, let me change that

clinicadl/data/datatype/modalities/__init__.py
thibaultdvx: Changes made, thanks for the suggestion

clinicadl/data/tensor_conversion.py
thibaultdvx: With `mask=Mask('caps/masks/leftHippocampus.nii.gz')`, `mask.name` is `leftHippocampus` and `mask.path.name` is `leftHippocampus.nii.gz`. Here it is better to have the full name to identify the mask.

PR #700, opened 2025-02-18
[DRAFT] First try for the Trainer

camillebrianceau:
In this PR (feel free to change whatever you want, this PR is just a good start for the Modules we need):

- `API/complicated_case.py`: API model with the correct use of each object.
- `BatchLoader`: A new kind of dataloader that is returned during training for each batch (`for batch, batchloader in enumerate(train_loader)`).
- Add a `MapsReader`: This object's goal is ONLY to read/return paths. Many methods need to be moved to another object.
- Add an `Optimum` attribute for all metrics to calculate the best model (need to know if minimize or maximize).
- Add a `get_datapoint` method to the `extraction` object.
- Add an `id` attribute to the `extraction` object to save and read in a TSV file (slice position or patch index).
- Create `OutputTransform`: Needs improvement but will be almost the same as `Transform`, but for the output.
- Add some methods in `Trainer`, `Predictor`, `Metrics`, `ClinicaDLModel`, which will be the main objects.
- Add a `maps_test` example of output for now.


clinicadl/utils/computational/computational.py
```diff
@@ -1,19 +1,21 @@
...
-class ComputationalConfig(BaseModel):
+class ComputationalConfig(ClinicaDLConfig):
     """Config class to handle computational parameters."""
 
     amp: bool = False
     fully_sharded_data_parallel: bool = False
-    gpu: bool = True
+    gpu: bool = False
     # pydantic config
     model_config = ConfigDict(validate_assignment=True)
```
thibaultdvx:
```suggestion
```

clinicadl/utils/computational/computational.py
```diff
@@ -31,3 +33,7 @@ def check_gpu(self) -> Self:
                 "AMP is designed to work with modern GPUs. Please add the --gpu flag."
```
thibaultdvx: I think amp also works with cpu. WDYT?

clinicadl/optim/config.py
```diff
@@ -2,16 +2,16 @@
...
 
-class OptimizationConfig(BaseModel):
+class OptimizationConfig(ClinicaDLConfig):
     """Config class to configure the optimization process."""
 
-    accumulation_steps: PositiveInt = 1
+    accumulation_steps: PositiveInt = 1  # gives the number of iterations during which gradients are accumulated before performing the weights update. This allows to virtually increase the size of the batch. Default: 1.
+    evaluation_steps: PositiveInt = 5  # gives the number of iterations to perform an evaluation internal to an epoch. Default will only perform an evaluation at the end of each epoch.
     epochs: PositiveInt = 10
+    non_blocking: bool = True
```
thibaultdvx: I find it more relevant in `ComputationalConfig`. WDYT?

clinicadl/optim/optimizers/config.py
```diff
@@ -25,6 +25,9 @@
 ]
 
 
+# TODO: add momentum in Optimizer config momentum: float = 0.9
+
+
```
thibaultdvx: What do you mean exactly? There is already momentum in SGD for example

clinicadl/trainer/trainer.py
```diff
@@ -1,203 +1,393 @@
...
+
+        # METRICS CONFIG
+        metrics.set_computational(self.comp)
+        metrics.set_loss(model.loss)
+
+        # will be different if resume is called
+        self.current_epoch: int = 0
+
+        # seed initialization
+        seed_everything(123, deterministic=False, compensation="memory")
```
thibaultdvx, line 81: The seed could be a parameter passed by the user, WDYT?

clinicadl/trainer/trainer.py
```diff
@@ -1,203 +1,393 @@
...
-        model.load_optim_state_dict(
+        self.model.load_optim_state_dict(
             self.reader.optimizer_path(split.index, resume=True)
         )
-        current_epoch = model.load_state_dict(
+        self.current_epoch = self.model.load_state_dict(
             self.reader.checkpoint_path(split.index, resume=True)
         )
+        metrics = self.reader.load_metrics()
+
```
thibaultdvx, line 157: I think we also need to resume the LR scheduler. The [DIstributedSampler](https://pytorch.org/docs/stable/data.html#torch.utils.data.distributed.DistributedSampler) as well (to keep in mind).

clinicadl/trainer/trainer.py
```diff
@@ -1,203 +1,393 @@
...
+        ):
+            self.on_epoch_begin()
 
-                # If no evaluation has been performed, warn the user
-                elif evaluation_flag and evaluation_steps != 0:
-                    logger.warning(
-                        f"Your evaluation steps {evaluation_steps} are too big "
-                        f"compared to the size of the dataset. "
-                        f"The model is evaluated only once at the end epochs."
+            for batch, data in enumerate(split.train_loader):
```
thibaultdvx: I think `batch_idx` (or something like this) could be a bit more explicit.

clinicadl/trainer/trainer.py
```diff
@@ -1,203 +1,393 @@
...
+
+    def on_batch_end(self):
+        pass
+
+    def on_epoch_begin(self):
+        self.model.network.zero_grad(set_to_none=True)
+        # self.evaluation_flag = True
+
+    def on_epoch_end(self, split: Split):
+        self.model.network.zero_grad(set_to_none=True)
```
thibaultdvx: Already done in `on_epoch_begin`, no?

clinicadl/data/dataloader/batchloader.py
```diff
@@ -0,0 +1,88 @@
+import torch
+
+from clinicadl.transforms.extraction import Sample
+
+
+class BatchLoader:
```
thibaultdvx: I'm opening a debate: is it really a `loader`? Personally, I would rather call it `Batch` or `Samples` or `BatchSamples`. WDYT?

clinicadl/data/dataloader/batchloader.py
```diff
@@ -0,0 +1,88 @@
...
+        ------
+            ValueError: If the provided list of samples is empty.
+        """
+        # Store the samples of the batch
+        self.samples = samples
+
+        if len(self) == 0:
+            raise ValueError("No samples to load.")
+
+    def __len__(self) -> int:
```
thibaultdvx: If you inherit from `list`, you will have `__len__` and `__getitem__` for free:

```
class BatchLoader(list):

    def __init__(self, samples: list[Sample]):
        super().__init__(samples)

        if len(self) == 0:
            raise ValueError("No samples to load.")

    def get_images(self) -> torch.Tensor:
        # Return the images of the batch
        return torch.cat([sample.sample for sample in self], dim=0).unsqueeze(1)

    def get_labels(self) -> torch.Tensor:
        # Return the labels of the batch
        if all(isinstance(sample.label, torch.Tensor) for sample in self):
            list_ = []
            for sample in self:
                list_.append(sample.label)
            return torch.cat(list_, dim=0).unsqueeze(1)
        else:
            return torch.tensor(
                [sample.label for sample in self], dtype=torch.float32
            ).unsqueeze(1)  # TODO: check torch.long
```

clinicadl/data/dataloader/batchloader.py
```diff
@@ -0,0 +1,88 @@
...
+        -------
+            torch.Tensor: A tensor containing all the labels from the batch.
+        """
+        # Return the labels of the batch
+        if all(isinstance(sample.label, torch.Tensor) for sample in self.samples):
+            list_ = []
+            for sample in self.samples:
+                list_.append(sample.label)
+            return torch.cat(list_, dim=0).unsqueeze(1)
+        else:
```
thibaultdvx, line 58: There is an issue for reconstruction, i.e. when label is `None`.

clinicadl/data/dataloader/batchloader.py
```diff
@@ -0,0 +1,88 @@
...
+    def get_images(self) -> torch.Tensor:
+        """
+        Get the images from the samples in the batch.
+
+        Returns
+        -------
+            torch.Tensor: A tensor containing all the images from the batch.
+        """
+        # Return the images of the batch
+        return torch.cat([sample.sample for sample in self.samples], dim=0).unsqueeze(1)
```
thibaultdvx: Why `unsqueeze`? There is already a channel dimension in the sample.

clinicadl/data/dataloader/batchloader.py
```diff
@@ -0,0 +1,88 @@
...
+        # Return the labels of the batch
+        if all(isinstance(sample.label, torch.Tensor) for sample in self.samples):
+            list_ = []
+            for sample in self.samples:
+                list_.append(sample.label)
+            return torch.cat(list_, dim=0).unsqueeze(1)
+        else:
+            return torch.tensor(
+                [sample.label for sample in self.samples], dtype=torch.float32
+            ).unsqueeze(1)  # TODO: check torch.long
```
thibaultdvx:
```suggestion
            return torch.tensor(
                [sample.label for sample in self.samples if sample.label else sample.sample], dtype=torch.float32
            ).unsqueeze(1)  # TODO: check torch.long
```

clinicadl/trainer/trainer.py
```diff
@@ -1,203 +1,393 @@
...
+        # Update learning rate based on validation loss
+        validator = Predictor(self.reader.maps_path, self.model, self.comp)
```
thibaultdvx, line 211: Why is the predictor initialized here?

clinicadl/metrics/metrics.py
```diff
@@ -0,0 +1,299 @@
...
+ReconstructionLosses = [
+    "L1Loss",
+    "MSELoss",
+    "KLDivLoss",
+    "BCEWithLogitsLoss",
+    "HuberLoss",
+    "SmoothL1Loss",
+    "VAEGaussianLoss",
+    "VAEBernoulliLoss",
+    "VAEContinuousBernoulliLoss",
```
thibaultdvx: Outdated

thibaultdvx: What are those lists for?

clinicadl/metrics/metrics.py
```diff
@@ -0,0 +1,299 @@
...
+
+        self.df = self._init_df()
+
+    def get_loss(self, epoch: Optional[int] = None):
+        if epoch:
+            return self.df.at[epoch, LOSS]
+        else:
+            return self.df[LOSS].iloc[-1]
+
+    def aggregate(self, epoch: int = 0):
```
thibaultdvx, line 63: I think it would be nice if that object has a `__call__` function. This function would call the computation of each metric. We would then do:

```
metrics(outputs, labels)
```

instead of:

```
for callable_metric in metrics._callable_metrics.values():
    callable_metric(outputs, labels)
```

clinicadl/metrics/metrics.py
```diff
@@ -0,0 +1,299 @@
...
+            metrics_list.append(metric)
+
+        if not all(isinstance(metric, ImplementedMetric) for metric in metrics_list):
+            raise ValueError(
+                "All metrics must be implemented in ClinicaDL in order to be used."
+            )
+
+        return metrics_list
+
+    def set_computational(self, computational_config: ComputationalConfig):
```
thibaultdvx: Why do we need that? Metrics computation is always do classically on a cpu, isn't it?

clinicadl/metrics/metrics.py
```diff
@@ -0,0 +1,299 @@
...
+    def model_dump(self):
+        dict_ = {}
+        dict_["metrics"] = {
+            "metrics": self.metrics,
+            "selection_metrics": self.selection_metrics,
+        }
+        return dict_
+
+
+class Metrics:
```
thibaultdvx, line 147: I understand why you created this object but I'm not certain about its usefulness. Moreover, it creates some confusion with `GroupMetrics`.

So, I think we can manage without that object. Or we could create a more general object gathering all kinds of logs (metrics, computation times, tensorboard logs, etc.) in the same place.

clinicadl/metrics/metrics.py
```diff
@@ -0,0 +1,299 @@
...
+    def get_value(self, epoch: int, metric: str):
+        try:
+            return self.df.at[epoch, metric]
+        except KeyError:
+            raise KeyError(f"Metric '{metric}' not found in the provided metrics.")
+
+    @staticmethod
+    def check_metrics(
+        metrics: Union[MetricsTypes, list[MetricsTypes]],
+    ) -> list[ImplementedMetric]:
```
thibaultdvx, line 96: Ok so `check_metrics` returns a list of names of metrics, doesn't it? But if I passed a metric config class with some specific options, they will be lost, no?

clinicadl/metrics/metrics.py
```diff
@@ -0,0 +1,299 @@
...
+    "BCEWithLogitsLoss",
+    "HuberLoss",
+    "SmoothL1Loss",
+]
+
+
+class GroupMetrics:
+    def __init__(
+        self,
+        metrics: Union[MetricsTypes, list[MetricsTypes]],
```
thibaultdvx, line 40: Personally, I would only allow `MonaiMetric` or `MetricConfig` for the inputs.

clinicadl/predictor/predictor.py
```diff
@@ -1,71 +1,296 @@
...
+        else:
+            self.model = model
+
+    def validate(
+        self,
+        dataloader: DataLoader,
+        metrics: GroupMetrics,
+        epoch: int = 0,
+    ):
+        self.model.network.eval()
```
thibaultdvx, line 80: I think this is where we should call `dataloader.dataset.eval()` as well, to have the CapsDataset in eval mode (to avoid applying data augmentation).

clinicadl/trainer/trainer.py
```diff
@@ -1,203 +1,393 @@
...
+
+                images = data.get_images().to(self.comp.device)
+                labels = data.get_labels().to(self.comp.device)
+
+                with autocast(self.comp.device.type, enabled=self.comp.amp):
+                    outputs = self.model.network(images)
+                    loss = self.model.loss(outputs, labels)
+
+                    self.metrics.write_training_loss(
+                        epoch=self.epoch, batch=batch, loss=loss.item()
```
thibaultdvx: I would typically put this in the `ClinicaDLModel` for more flexibility. I'm thinking of a method named `training_step`:

```
def training_step(self, data: BatchLoader) -> torch.Tensor:
    """
    Takes a batch as input and return the computed loss.
    The user can overwrite this default implementation.
    """
    images = data.get_images().to(self.comp.device)
    labels = data.get_labels().to(self.comp.device)
    outputs = self.network(images)
    loss = self.loss(outputs, labels)

    return loss
```

So, `Trainer.train` would then look like this:

```
def train(self, split: Split):
    """TO COMPLETE"""

    self.on_train_begin(split)

    while self.epoch < self.optim.epochs and not self.early_stopping.step(
        self.metrics.val.get_loss()
    ):
        self.on_epoch_begin()

        for batch, data in enumerate(split.train_loader):
            self.on_batch_begin()

            with autocast(self.comp.device.type, enabled=self.comp.amp):
                loss = self.model.training_step(data)

            self.metrics.write_training_loss(
                epoch=self.epoch, batch=batch, loss=loss.item()
            )

            self.scaler.scale(loss).backward()
            self.weights_update()

        self.on_epoch_end(split)

    self.on_train_end(split)
```

thibaultdvx: This would not work because of `.to(self.comp.device)` in `training_step`, but you get the idea.

clinicadl/utils/early_stopping.py
```diff
@@ -0,0 +1,43 @@
+class EarlyStopping(object):
```
thibaultdvx: To compare with the `EarlyStopping` already implemented in `optim/early_stopping/early_stopper.py`, but they look really similar.

clinicadl/predictor/predictor.py
```diff
@@ -1,71 +1,296 @@
...
+        with torch.no_grad():
+            for batch, data in enumerate(dataloader):
+                ############
+                images = data.get_images().to(self.comp.device)
+                labels = data.get_labels().to(self.comp.device)
+                ############
+
+                # initialize the loss list to save the loss components
+                with autocast(self.comp.device.type, enabled=self.comp.amp):
+                    outputs = self.model.network(images)
```
thibaultdvx, line 94: Here as well, I think it would be more flexible to add a `predict` method in the `ClinicaDLModel`. WDYT?

clinicadl/metrics/metrics.py
```diff
@@ -0,0 +1,299 @@
...
+            return self.df.at[epoch, LOSS]
+        else:
+            return self.df[LOSS].iloc[-1]
+
+    def aggregate(self, epoch: int = 0):
+        for metric, callable_metric in self._callable_metrics.items():
+            value = callable_metric.aggregate()
+            self.df.at[epoch, metric] = value.item()
+
+    def _reset_callable(self):
```
thibaultdvx: Maybe just `reset`? To be consistent with MONAI's nomenclature.

clinicadl/trainer/trainer.py
camillebrianceau: Because the predictor is initialized from a model with updated weights?

camillebrianceau:
Thanks for your review, @thibaultdvx! I've implemented most of your suggestions and agree with almost all of your comments. However, I think the `Metrics` module requires significant changes, so it might be better to address that in a future PR. The same applies to the `BatchLoader` and some other changes. Since this PR is already quite large, I was thinking we could merge it as a draft to serve as a foundation for our upcoming modifications.  

### Summary of planned changes  

#### Refactor `Metrics` module  (PR #706)
- Only allow `MonaiMetric` or `MetricConfig` as inputs  
- Add a `__call__` method  
- Modify the list of metric names argument to prevent losing specific metric options  
- Remove `GroupMetric`, as it is not particularly useful  

#### Refactor `BatchLoader`  (PR #707)
- Consider renaming it (e.g., `BatchSamples`)  
- Check if `unsqueeze` is necessary, given that the sample already has a channel dimension  
- Find a way to handle cases where the label is `None` (for reconstruction tasks)  

#### Enhance `ClinicaDLModel`  PR #708)
- Add `training_step` and `predict` methods for more flexibility  

Let me know what you think!

clinicadl/splitter/split.py
```diff
@@ -134,6 +134,8 @@ def build_train_loader(
         KeyError
             If the column passed in `sampling_weights` cannot be converted to floats.
         """
+
+        self.train_dataset._count_samples()
```
thibaultdvx, line 138: Why do we need this line?

clinicadl/splitter/split.py
```diff
@@ -152,6 +154,7 @@ def build_train_loader(
             dp_degree=self._dp_degree,
             rank=self._rank,
         )
+        self.train_loader.type = "train"
```
thibaultdvx: This will raise an error, no? A PyTorch DataLoader doesn't have an attribute named `type`.

clinicadl/transforms/output_transforms.py
```diff
@@ -0,0 +1,210 @@
...
+    def apply(
+        self,
+        data_point: DataPoint,
+    ) -> DataPoint:
+        output = self.get_transforms()(data_point)
+
+        if data_point.image.shape != output.image.shape:
+            raise ValueError(
+                "One of the transform modifies the size of the tensor and can't be used"
+            )
```
thibaultdvx: I'm just wondering why it is a problem to modify the size of the output in the postprocessing.

thibaultdvx (approved):

thibaultdvx:
LGTM @camillebrianceau, thanks for the great work!
Just left some last comments and committed some modifications:
- add tests for the handling of the labels in CapsDataset;
- correct some metrics optima that were wrong (tell me if you don't agree). Particularly, we need to be careful with `ConfusionMatrixMetric` since it gathers a lot of different metrics;
- simplify a bit `OutputTransforms`.

Feel free to merge, and we will make the changes you mention later!

clinicadl/splitter/split.py
camillebrianceau: I don't know if it is the right place to do that, but if we don't have this line, the dataset at the validation step or the beginning of the second split would have an empty df. (Or maybe I am missing something about how slice indexes are used...)
You can try to run the `complicated_case. py ' for an example.

clinicadl/splitter/split.py
camillebrianceau: Yes !! It was a test and I forgot to remove that part !!

clinicadl/transforms/output_transforms.py
camillebrianceau: It is not 🤔

PR #702, opened 2025-02-21
Api doc

thibaultdvx:
First draft for API documentation using Sphinx:
https://clinicadl--702.org.readthedocs.build/en/702/index.html


.readthedocs.yaml
```diff
@@ -18,7 +18,7 @@ build:
       # Do not use virtual environments
       - poetry config virtualenvs.create false
       # Install docs group of dependencies
-      - poetry install --only docs
+      - poetry install --with=docs
 
 mkdocs:
```
NicolasGensollen:
```suggestion
sphinx:
```

thibaultdvx: Works better indeed, thanks @NicolasGensollen ;)

clinicadl/transforms/config/intensity.py
```diff
@@ -147,7 +170,10 @@ def _get_class(self) -> type[tio.Transform]:
 
 
 class ClampConfig(TransformConfig):
-    """Config class for Clamp transform."""
+    """
+    Config class for TorchIO's `Clamp <https://torchio.readthedocs.io/transforms/preprocessing.html#torchio.transforms.Clamp>`_
```
NicolasGensollen: You should take a look at Sphinx [Intersphinx](https://www.sphinx-doc.org/fr/master/usage/extensions/intersphinx.html) functionality. The base idea is to write these URL only once for the projects that you depend on, and use a special sphinx syntax to cross-reference objects from these projects.
The huge advantage is that, if one day the URL change for some reason, you won't have to dig into hundreds of Python files to update those links.

thibaultdvx: Great idea, thanks!

camillebrianceau (approved):

PR #708, opened 2025-03-31
Maps Object

camillebrianceau:
Try something new for the MAPS reader: I created objects that represent each sub-folder of the MAPS architecture. Please let me know what you think! If you like it, I create the associated tests.


clinicadl/optim/config.py
```diff
@@ -14,3 +18,9 @@ class OptimizationConfig(ClinicaDLConfig):
     evaluation_steps: PositiveInt = 5  # gives the number of iterations to perform an evaluation internal to an epoch. Default will only perform an evaluation at the end of each epoch.
     epochs: PositiveInt = 10
     early_stopping: EarlyStoppingConfig = EarlyStoppingConfig()
+
+    def write_info(self, json_path: PathType):
+        update_json(json_path=Path(json_path), config=self)
```
thibaultdvx, line 23: Maybe we can put this method in ``ClinicaDLConfig``, as it useful for many config classes. WDYT?

clinicadl/maps/data_group/datagroup.py
```diff
@@ -0,0 +1,62 @@
...
+
+class BaseDataGroup(Directory):
+    @property
+    @abstractmethod
+    def name(self) -> str:
+        pass
+
+    @property
+    def data_tsv(self) -> Path:
+        return self.path / (DATA + TSV)
```
thibaultdvx, line 26:
```suggestion
        return (self.path / DATA).with_suffix(TSV)
```

Maybe a bit more explicit. WDYT?

thibaultdvx (approved):
LGTM, very elegant!

clinicadl/optim/config.py
camillebrianceau: Yes, but as I'm not sure how to save the information in the JSON files (all in one updated maps.json file or multiple files for each configuration class), I'd rather keep this one here for now.

thibaultdvx (approved):

PR #710, opened 2025-04-07
Transforms ZOO

camillebrianceau:
Introduce a first try for the Transforms Zoo. I added a tutorial to see how it works.


camillebrianceau:
Close because ZOO is actually in another repo https://github.com/aramis-lab/clinicadl-zoo/pull/1

PR #711, opened 2025-04-08
Read/Write json

camillebrianceau:
In This PR

- Renamed `utils.computational.computational` to `utils.computational.config` for better clarity and consistency.
- Removed the `experiment_manager` module, which is now replaced by the `maps` module.

- Added a `load` method to the `Maps` class, allowing loading of an existing `Maps` object.
- Propagated the `load` method to related classes: `SplitDir`, `DataGroup`, and `BestMetrics`.
- Created a new file `utils/json.py` to centralize all JSON-related methods:
   - `write_json`, `read_json`, `update_json`, `path_encoder`, and `path_decoder`.
- Added `from_json`, `write_json`, and `update_json` methods to the `ClinicaDLConfig` class for more flexible config handling.
- Implemented `from_json` and `write_json` methods in the `Metrics` class for easier serialization and deserialization.
- Introduced a `training_step` method in the `ClinicaDLModel` class as suggested.
- Clean a bit the `train` method in the `Trainer`


clinicadl/data/dataloader/batchloader.py
```diff
@@ -39,7 +39,9 @@ def get_images(self) -> torch.Tensor:
             torch.Tensor: A tensor containing all the images from the batch.
         """
         # Return the images of the batch
-        return torch.cat([sample.sample for sample in self], dim=0).unsqueeze(1)
+        return torch.cat(
+            [sample.get_tensors()["image"] for sample in self], dim=0
+        ).unsqueeze(1)
```
thibaultdvx:
```suggestion
        return torch.stack([sample.image.tensor for sample in self], dim=0)
```

``get_tensors`` will convert all the images to tensors. Here we only need one.

Besides, ``torch.stack`` is more suited than ``torch.cat``.

clinicadl/data/dataloader/batchloader.py
```diff
@@ -39,7 +39,9 @@ def get_images(self) -> torch.Tensor:
...
 
     def get_labels(self) -> torch.Tensor:
```
thibaultdvx: I think there are some issues with ``get_labels`` with ``None`` labels and heterogenous labels (e.g. in a ``ConcatDataset``). I would suggest something like that:

```python
def get_labels(self) -> Union[torch.Tensor, List[Any]]:
    """
    Gets the labels from the samples in the batch.

    Returns
    -------
    Union[torch.Tensor, List[Any]]
        A tensor or a list containing all the labels from the batch.
        It will be a list if the labels are heterogenous (e.g. a mask and a scalar) or if any
        of the label is ``None``. Otherwise, it will be a tensor.
    """
    labels = [
        sample.label.tensor if isinstance(sample.label, tio.LabelMap) else sample.label
        for sample in self
    ]
    if all(isinstance(label, torch.Tensor) for label in labels):
        return torch.stack(labels, dim=0)
    elif all(isinstance(label, (int, float)) for label in labels):
        return torch.tensor(
            labels,
            dtype=torch.float32,
        )
    else:
        return labels
```

thibaultdvx (approved):

thibaultdvx:
I think we are ready to merge. I just canceled the first merge with ``upstream/clinicadl_v2``, and did it again after correction of my mistake on ``clinicadl_v2``.

PR #713, opened 2025-04-15
Zoo

thibaultdvx:
- Small reorganisation in ``data.structures``.
- Add an example of a DataPoint (``ColinDataPoint``), useful for tutorials and examples.
- Add Zoo in dependencies. Via poetry, Zoo can be installed with ``poetry install --extras zoo``. Later, via pip, it will be ``pip install clinicadl[zoo]``.


clinicadl/data/structures/examples.py
camillebrianceau, line 1: Maybe we can put examples in the `test` folder WDYT?

camillebrianceau (approved):

clinicadl/data/structures/examples.py
thibaultdvx: If the user install clinicadl with pip, (s)he won't have access to them if they are in ``tests``.

PR #715, opened 2025-04-18
Remove case insensitive glob

thibaultdvx:
Remove [case insensitive glob](https://github.com/aramis-lab/clinicadl/blob/b6028ef0d2a6938dbf6403c905ffb999542429c3/clinicadl/data/readers/utils.py#L4) in ``CapsReader`` as it is not really useful and leads to some bugs:
https://stackoverflow.com/questions/8151300/ignore-case-in-glob-on-linux

Additionally, there were some issues when reading json in ``TensorConversion``, so I fixed them by using ``from_json`` classmethod of ``ClinicaDLConfig``.


camillebrianceau (approved):
LGTM !

PR #717, opened 2025-04-24
Trainer + Metrics + Callbacks

camillebrianceau:
[DRAFT]

- Refactored the `metrics` module.
- Restored the previous `callbacks` module with an initial list of functional examples:
  - `Chronometer` – working
  - `CodeCarbon` – needs update
  - `Logger` – working
  - `MLflow` – needs update
  - `WandB` – needs update
- Added a `plot_loss()` method to the `Maps` object, but don't know if it is a useful feature, WDYT?
  → Example usage: `maps.split[1].logs.plot_loss()`
- Proposed converting **EarlyStopping** into a callback , WDYT ?
- Removed `training_config` and other unused/obsolete files for cleanup.

Suggestions:

- Would love your thoughts on the current changes.
- I believe we can expand the callback system further for a cleaner, more modular trainer.
  - *Maybe even move metrics into callbacks for better separation of concerns?*


clinicadl/callbacks/factory/chronometer.py
```diff
@@ -0,0 +1,244 @@
...
+    """
+    A lightweight profiler for timing a PyTorch training loop.
+
+    Tracks durations of training phases like data loading, forward pass,
+    backward pass, and validation. Supports summary display and export.
+
+    Example
+    -------
+    chrono = Chronometer()
+    chrono.start()
```
thibaultdvx: I don't see the method ``start``. Where is it?

clinicadl/callbacks/factory/chronometer.py
```diff
@@ -0,0 +1,244 @@
...
+    def clear(self) -> None:
+        """Clears all recorded timing data."""
+        self.time_perf_train.clear()
+        self.time_perf_load.clear()
+        self.time_perf_forward.clear()
+        self.time_perf_backward.clear()
+
+    def elapsed(self) -> Optional[float]:
+        """
+        Returns time elapsed since `start()` was called.
```
thibaultdvx: Since ``on_train_begin`` was called?

clinicadl/callbacks/factory/chronometer.py
```diff
@@ -0,0 +1,244 @@
...
+    Tracks durations of training phases like data loading, forward pass,
+    backward pass, and validation. Supports summary display and export.
+
+    Example
+    -------
+    chrono = Chronometer()
+    chrono.start()
+
+    for epoch in range(epochs):
+        chrono.next_iter()
```
thibaultdvx: So ``next_iter`` is called two times in a row?

clinicadl/callbacks/base.py
```diff
@@ -0,0 +1,200 @@
...
+        callback : Callback
+            Callback instance to be added.
+        """
+        if isinstance(callback, type):
+            for cb in self.callbacks:
+                if isinstance(cb, callback):
+                    self.callbacks.remove(cb)
+                    return
+        else:
+            self.callbacks.remove(callback)
```
thibaultdvx: Not sure to understand this method.

Type hinting suggests that I can do:

```python
from clinicadl.callbacks import CallbacksHandler
from clinicadl.callbacks.factory import LoggerCallback
callback_handler = CallbacksHandler(logger=True)
callback_handler.remove_callback(LoggerCallback())
```

Right?

But this raises the error: `ValueError: list.remove(x): x not in list`

Besides I don't know if this method will be very useful. WDYT?

clinicadl/callbacks/base.py
```diff
@@ -0,0 +1,200 @@
...
+    """
+
+    def __init__(
+        self,
+        mlflow: bool = False,
+        wandb: bool = False,
+        codecarbon: bool = True,
+        logger: bool = True,
+        chronometer: bool = True,
+        custom_callback: Optional[Union[Callback, list[Callback]]] = None,
```
thibaultdvx: A suggestion:
-  change to ``__init__(self, *args: Callback)`` here,
- handle the default values in Trainer's arguments:
   ```python
   class Trainer:
   ...
     def __init__(
       self,
       ...
       callbacks=[CodeCarbonCallback(), LoggerCallback(), Chronometer(), ...],
     ):
   ```

The user then manipulates ``Callbacks`` and not ``CallbacksHandler``.

Tell me what do you think.

clinicadl/metrics/enum.py
```diff
@@ -0,0 +1,36 @@
+from enum import Enum
+
+
+class ClassificationMetrics(Enum):
```
thibaultdvx: Where is used this class?

clinicadl/metrics/metrics.py
```diff
@@ -1,258 +1,304 @@
...
+        """
+        if isinstance(v, MetricConfig):
+            return {v.name: v}
+        elif isinstance(v, list):
+            metrics = {}
+            for metric in v:
+                if not isinstance(metric, MetricConfig):
+                    raise TypeError(f"Expected MetricConfig, got {type(metric)}")
+                if metric.name in metrics:
+                    raise ValueError(f"Duplicate metric name '{metric.name}' found.")
```
thibaultdvx: Maybe the user will want to use the same metric with different parameters.

clinicadl/metrics/metrics.py
```diff
@@ -1,258 +1,304 @@
...
+        for name, metric in self._callable_metrics.items():
+            value = metric.aggregate().item()
+            if batch is not None:
+                self._df.at[(epoch, batch), name] = value
+            else:
+                self._df.at[epoch, name] = value
 
-    @classmethod
-    def from_json(cls, json_path: Path) -> Metrics:
+    def __call__(self, y_pred: torch.Tensor, y: torch.Tensor) -> None:
```
thibaultdvx:
```suggestion
    def __call__(self, y_pred: torch.Tensor, y: Optional[torch.Tensor] = None, **kwargs) -> None:
```

To match: 

https://github.com/Project-MONAI/MONAI/blob/46a5272196a6c2590ca2589029eed8e4d56ff008/monai/metrics/metric.py#L327

clinicadl/metrics/metrics.py
```diff
@@ -1,258 +1,304 @@
...
+            Flag to compute training metrics.
+        """
+        self.metrics: Dict[str, MetricConfig] = self._check_list_metrics(metrics)
+        self.selection_metrics: Optional[Dict[str, MetricConfig]] = (
+            self._check_list_metrics(selection_metrics) if selection_metrics else None
+        )
 
-        self.selection_metrics = self.check_metrics(selection_metrics)
+        self.compute_train_metrics = compute_train_metrics
+        self._df = pd.DataFrame(
```
thibaultdvx: I would put ``_df`` public. As it sumarizes all the metrics, it may be a useful attribute to access for the user/developper. WDYT?

clinicadl/metrics/metrics.py
```diff
@@ -1,258 +1,304 @@
...
 
-    def model_dump(self):
-        return self.val.model_dump()
+        Returns
+        -------
+        dict
+            Serialized metrics and selection_metrics.
+        """
+        return {
+            METRICS: [metric.model_dump() for metric in self.metrics.values()],
```
thibaultdvx:
```suggestion
            METRICS: [metric.to_dict() for metric in self.metrics.values()],
```

clinicadl/metrics/metrics.py
```diff
@@ -1,258 +1,304 @@
...
+            SELECTION_METRICS: (
+                [metric.model_dump() for metric in self.selection_metrics.values()]
```
thibaultdvx:
```suggestion
                [metric.to_dict() for metric in self.selection_metrics.values()]
```

clinicadl/metrics/metrics.py
```diff
@@ -1,258 +1,304 @@
...
-        self.val.set_loss(loss)
+        Parameters
+        ----------
+        best_metrics : Dict[str, BestMetric]
+            Mapping of metric names to their best-tracking wrappers.
+        """
+        if not self.selection_metrics:
+            raise RuntimeError("Cannot save metrics without selection_metrics.")
+        for name in self.selection_metrics:
+            self._df.to_csv(best_metrics[name].val.metrics_tsv, sep="\t", index=True)
```
thibaultdvx: Sorry but I don't understand where is saved the metric dataframe. Why always in the validation folder?

clinicadl/metrics/metrics.py
```diff
@@ -1,258 +1,304 @@
...
+        KeyError
+            If metric name is invalid.
+        """
+        if name not in self._df.columns:
+            raise KeyError(
+                f"Metric '{name}' not found. Available: {self._df.columns.tolist()}"
+            )
+        if epoch not in self._df.index:
+            return np.nan
+        return self._df.at[epoch, name]
```
thibaultdvx: Really useful? ``metrics.df.at[epoch, name]`` is quite straightforward, no?

clinicadl/metrics/metrics.py
```diff
@@ -1,258 +1,304 @@
...
-        self.metrics = self.check_metrics(metrics)
-
-        if ImplementedMetric.LOSS not in self.metrics:
-            self.metrics.append(ImplementedMetric.LOSS)
+        """
+        Initialize the ClinicaDLMetrics instance.
+
+        Parameters
+        ----------
+        metrics : MetricType
```
thibaultdvx, line 85: Can this object also accepts custom metrics, like the ``Transforms`` object?

clinicadl/metrics/metrics.py
```diff
@@ -1,258 +1,304 @@
...
+
+        Parameters
+        ----------
+        epoch : int
+            Current epoch.
+        batch : Optional[int]
+            Current batch (optional).
+        """
+        for name, metric in self._callable_metrics.items():
+            value = metric.aggregate().item()
```
thibaultdvx: Just to be sure: when you do ``aggregate``, you take the mean from all the batches of all the passed epochs if you don't reset the metric at the beginning of each epoch.

So, I'm not sure this is what you want to put in the DataFrame here, is it?

thibaultdvx: And there are (at least) two tricky cases with ``aggregate``:  ``ROCAUCMetric`` and ``ConfusionMatrixMetric`` don't return a tensor.

```python
>>> import monai.metrics as metrics
>>> roc_auc = metrics.ROCAUCMetric()
>>> precision = metrics.ConfusionMatrixMetric(metric_name="precision")
>>> x = torch.randint(0, 2, (5, 3))
>>> y = torch.randint(0, 2, (5, 3))
>>> roc_auc(x, y)
>>> roc_auc.aggregate()
0.3888888888888889
>>> precision(x, y)
>>> precision.aggregate()
[tensor([0.4286])]
```

clinicadl/metrics/metrics.py
```diff
@@ -1,258 +1,304 @@
...
-        return cls.from_dict(dict_)
+        for metric in self._callable_metrics.values():
+            metric(y_pred, y)
 
-    @classmethod
-    def from_dict(cls, dict_: dict):
-        metrics_config = dict_["metrics"]
-        metrics = metrics_config["metrics"]
-        selection_metrics = metrics_config["selection_metrics"]
+    def get_loss(self) -> float:
```
thibaultdvx, line 223: When this method will be used?

clinicadl/optim/early_stopping/early_stopper.py
```diff
@@ -2,10 +2,12 @@
 
 import numpy as np
 
+from clinicadl.callbacks.base import Callback
+
 from .config import EarlyStoppingConfig, Mode
 
 
-class EarlyStopping(object):
+class EarlyStopping(Callback):
```
thibaultdvx: It is nice as a Callback indeed! So what do you think of moving this class in the folder of the callbacks?

clinicadl/callbacks/factory/chronometer.py
```diff
@@ -0,0 +1,244 @@
...
+from typing import List, Optional
+
+import numpy as np
+
+from .base import Callback
+
+logger = getLogger("clinicadl.chronometer")
+
+
+class Chronometer(Callback):
```
thibaultdvx, line 15: Do you have an example using this class? I think I'm missing some things

clinicadl/maps/split_dir/split_dir.py
```diff
@@ -154,6 +161,51 @@ def tensorboard(self) -> Path:
     def training_tsv(self) -> Path:
         return (self.path / TRAINING).with_suffix(TSV)
 
+    def plot_loss(self):
```
thibaultdvx, line 171: Do you have an example of such a plot?

clinicadl/data/dataloader/batch.py
```diff
@@ -41,6 +41,7 @@ def get_images(self) -> Union[torch.Tensor, list[torch.Tensor]]:
             dimension.
         """
         images = [sample.image.tensor for sample in self]
+        # images = [sample.image.tensor.squeeze(0) for sample in self]
```
thibaultdvx: Why did you comment?

thibaultdvx (commented):
Hi @camillebrianceau, thanks for the huge work!
Lots of comments. As always, feel free to only pick what you think relevant!

camillebrianceau:
This PR is finally ready and turned out to be larger than expected — sorry for that.

### Callbacks

Restored the previous `callbacks` module with an initial list of functional examples:

- `Chronometer`: working
- `CodeCarbon`: needs update
- `Comet`: to be implemented
- `EarlyStopping`: currently working, may require improvements
- `LearningScheduler`: to be implemented
- `Logger`: working
- `MLflow`: needs update
- `ModelCheckpoint`: currently working, may require improvements
- `ProgressBar`: needs update
- `TensorBoard`: to be implemented
- `WandB`: needs update

### Core Changes

- Replaced `stack` with `cat` in `data.dataloader.batch.SimpleBatch.get_images()` to prevent the addition of an extra dimension.
- Added a `plot_loss()` method to the `Maps` object.
  - This introduces a `matplotlib` dependency.
  - Might be helpful to provide built-in tools for inspecting content directly from the Maps object.
  - Example usage: `maps.split[1].logs.plot_loss()`

### Metrics

- Added docstrings to all metric classes for improved clarity and IDE support (e.g., VSCode).
- Added `MetricConfig` classes inheriting from `ConfusionMatrixMetricConfig`.
  - Not yet integrated into the factory; this will be addressed in a future PR.
- Added `CustomMetricConfig` to allow users to define their own metrics by inheriting from this class.
- Introduced `ClinicaDLMetrics`, a new component that is only used internally by the trainer.
  - Users now only provide a list of metrics to the trainer.

### Trainer & Predictor

- First draft of the `Predictor` class with a working `test()` method (name subject to change).
- Removed `training_config` and other obsolete or unused files.
- Implemented `_TrainingConfig`, a config class designed to pass useful arguments from the trainer to callbacks. WDYT?
- First draft of the `Trainer` class is functional.
  - Still needs cleanup and docstrings.
  - WDYT of the architecture?

### Next Steps

I will add all necessary tests and docstrings after the sprint next week.

clinicadl/callbacks/factory/early_stopping.py
```diff
@@ -0,0 +1,174 @@
...
+
+            if self.is_better(value, self.best):
+                self.num_bad_epochs = 0
+                self.best = value
+            else:
+                self.num_bad_epochs += 1
+
+            if self.num_bad_epochs >= self.patience:
+                return True
+
```
thibaultdvx, line 112:
```suggestion
            if self.num_bad_epochs >= self.patience:
                return True
                
            return False
```

Tell me if I'm wrong, but I think we are missing a last ``return False``, if none of the condition is met.

clinicadl/data/dataloader/batch.py
```diff
@@ -42,7 +42,7 @@ def get_images(self) -> Union[torch.Tensor, list[torch.Tensor]]:
         """
         images = [sample.image.tensor for sample in self]
         try:
-            return torch.stack(images, dim=0)
+            return torch.cat(images, dim=0)
```
thibaultdvx: I don't think this is right. You are squeezing the channel dimension. Even if it is almost always 1 in our cases, the channel dimension is essential to do convolution.

clinicadl/utils/config/training.py
```diff
@@ -0,0 +1,27 @@
...
+from clinicadl.metrics.metrics import ClinicaDLMetrics
+from clinicadl.model.clinicadl_model import ClinicaDLModel
+from clinicadl.optim.config import OptimizationConfig
+from clinicadl.splitter.split import Split
+from clinicadl.utils.computational.config import ComputationalConfig
+
+from .base import ClinicaDLConfig
+
+
+class _TrainingConfig(ClinicaDLConfig):
```
thibaultdvx: I think I understand the purpose of this class, but I just find the name a bit confusing. Maybe `_TrainingState` or something like that?

thibaultdvx: Maybe what I also find confusing is that you have in that class some attributes that are dynamic (e.g. epoch or batch) and some that are config classes.

But we can discuss it later.

clinicadl/metrics/config/base.py
```diff
@@ -53,6 +53,12 @@ class _ReductionConfig(ClinicaDLConfig):
 
     reduction: Reduction
 
+    @field_validator("reduction", mode="after")
+    @classmethod
+    def validator_reduction(cls, v):
+        if v == "mean_batch":
+            return Reduction.MEAN
+
```
thibaultdvx: I don't understand the purpose of this validator. ``Reduction`` only accepts ``mean`` and ``sum``, so it will never be ``mean_batch``.

clinicadl/metrics/config/segmentation.py
```diff
@@ -114,9 +123,13 @@ class GeneralizedDiceScoreConfig(MetricConfig, _IncludeBackgroundConfig):
     def __init__(
         self,
         include_background: Union[bool, DefaultFromLibrary] = DefaultFromLibrary.YES,
-        reduction: Union[Reduction] = Reduction.MEAN,
+        reduction: Union[Reduction, DefaultFromLibrary] = DefaultFromLibrary.YES,
```
thibaultdvx: In MONAI versions 1.4.*, the [default](https://docs.monai.io/en/stable/metrics.html#generalized-dice-score) for ``reduction`` in ``GeneralizedDiceScoreConfig`` is ``mean_batch``, so this is not what we want (we only want ``mean`` or ``sum``). Maybe you were looking at the 1.5.0.

Anyways, we can deal with that in another PR. Besides, I think it's a good idea to switch to MONAI 1.5.0, as it is compatible with numpy 2.*.

thibaultdvx (commented):
Thanks!!
Just few comments, but I think we can merge and talk about it later. Can you please juste have a look at the comment on ``.cat`` and the comment on early stopping?

PR #719, opened 2025-05-19
Splitter

thibaultdvx:
The main goal of this PR is to make the ``splitter`` module ready to use. To do that, I did:
- adapt the splitters to the new ``ConcatDataset``, ``PairedDataset``, and ``UnpairedDataset``;
- enhance docstrings;
- create pages on the splitters in the documentation;
- update test data, to be consistent with the other tests;
- complete unittests.

I also did:
- change a bit the ``subset`` function of ``CapsDataset`` (and its derivatives), so that it doesn't raise an error if a (participant, session) couple in the input DataFrame is not in the datasets;
- remove unused functions from ``split_utils.py``, and put this file in ``make_splits`` because it only concerns ``make_split`` and ``make_kfold``;
- rewrite some tsv tools like ``read_data`` (function to read a tsv) in ``tsvtools/utils.py``, and remove unused functions in this python file;
- remove default values from ``SingleSplitConfig`` and ``KFoldConfig`` to only keep them in ``make_split`` and ``make_kfold``;
- put the checks made on the split directories directly in ``SingleSplitConfig`` and ``KFoldConfig``;
- reduce docstrings of private functions in order to have smaller files;
- reorganize files to have always the main function at the top.

Besides, **I would particularly want to have your opinion on two changes I made**:
- for ``SingleSplit``, I suggested to change ``get_splits`` to ``get_split``, because I found it confusing to have plural here;
- I also changed ``validate_longitudinal`` to ``longitudinal``.


thibaultdvx:
To fix multiprocessing issues with ``num_workers`` for Python 3.12 on MacOS, I made pytest skip the multiprocessing tests on MacOS machines.

camillebrianceau (approved):
LGTM! Let's merge it so I can rebase before Monday! 

And you can change `get_splits` to `get_split` if you prefer.

PR #722, opened 2025-06-18
Add default value to the doc 

camillebrianceau:
Changes included in this PR:
- Renamed `splitter` folder to `split`
- Removed `DefaultFromLibrary` and replaced it with a default directory argument in each Config class for:
    - Metrics
    - Optimizers
    - LR Schedulers
    - Networks
    - Transforms
- Added ext and updated the documantation (typing)
- Adapted and updated related tests


clinicadl/split/make_splits/single_split.py
```diff
@@ -75,7 +75,7 @@ def make_split(
             Here, we are talking about number of **participants**. So, if ``n_test=0.2``, it doesn't mean that you have 80%
             of your data in the training set, but rather that you have 80% of you participants in the training set.
 
-    output_dir : Optional[Path, str], (optional, default=None)
+    output_dir : Union[Path, str], (optional, default=None)
```
thibaultdvx:
```suggestion
    output_dir : Optional[Union[Path, str]], (optional, default=None)
```

thibaultdvx (approved):
Thanks. Just added few changes.

PR #727, opened 2025-06-23
Trainer after sprint

camillebrianceau:
- change `trainer` into `train` directory
- remove `ProgressBar`callback and put it in `Logger`
- add `LRScheduler` as a callback
- add `validation_step`method to ClinicaDLModel


clinicadl/callbacks/factory/early_stopping.py
```diff
@@ -90,29 +91,29 @@ def on_epoch_end(self, config: _TrainingState, **kwargs):
...
 
-            if self.num_bad_epochs >= self.patience:
-                return True
+            if self.patience is not None and self.num_bad_epochs >= self.patience:
+                self.stop = True
 
-            return False
-
-        raise ValueError("No df provided")
+            self.stop = False
```
thibaultdvx, line 114: Not sure about this: there is no ``return`` above, so we will always arrive to this line, and thus ``self.stop`` will always be ``False``.

clinicadl/callbacks/factory/early_stopping.py
```diff
@@ -90,29 +91,29 @@ def on_epoch_end(self, config: _TrainingState, **kwargs):
...
-                return True
+                self.stop = True
 
             if self.lower_bound is not None and (value < self.lower_bound):
-                return True
+                self.stop = True
 
             if self.patience is None:
-                return False
+                self.stop = False
```
thibaultdvx, line 103: This can overwrite ``self.stop``  when it is ``True``. That was not the case before because of the ``returns``.

clinicadl/callbacks/factory/early_stopping.py
thibaultdvx: More generally, I think you should rethink the function because getting rid of ``returns`` changes all the logic.

clinicadl/callbacks/factory/early_stopping.py
```diff
@@ -172,5 +173,10 @@ def check_list(value) -> list:
             )
 
     def on_epoch_end(self, config: _TrainingState, **kwargs):
+        all_stop = True
         for metric in self.early_config_list:
             metric.on_epoch_end(config=config, **kwargs)
+            if not metric.stop:
+                all_stop = False
+
+        config.stop = all_stop
```
thibaultdvx:
```suggestion
    def on_epoch_end(self, config: _TrainingState, **kwargs):
        config.stop = all(metric.on_epoch_end(config=config, **kwargs) for metric in self.early_config_list)
```
And you keep all the ``returns`` in ``OneMetricEarlyStopping``. WDYT?

thibaultdvx (commented):
Thanks!
I think there are some issues with ``EarlyStopping``, otherwise LGTM!

clinicadl/callbacks/factory/early_stopping.py
camillebrianceau: Yes sorry I am changing all of this in the next PR !

PR #728, opened 2025-07-01
Callbacks Module

camillebrianceau:
CALLBACKS

**Private Callbacks** (internal use only — not meant to be provided by the user):
- `_CheckpointSaver`: Saves model and optimizer states at the end of each epoch in a temporary folder.
- `_Chronometer`: Tracks and logs time consumption for each training step.
- `_Logger`: Handles logging.
- `_TrainingLoss`: Stores the loss values throughout training.

**Public Callbacks** (provided by the user):
- `Checkpoint`: Saves model and optimizer at specified epochs or based on patience.
- `CodeCarbon`: Calculates CO2 emissions during training.
- `EarlyStopping`: Stops training when no improvement is detected.
- `LRScheduler`: Manages learning rate scheduling.
- `MLFLOW`: Logs experiments with MLflow.
- `ModelSelection`: Saves the best model according to a specified metric.
- `TensorBoard`: Integrates with TensorBoard.
- `WandB`: Integrates with Weights & Biases.

**Others**
- `CallbackHandler`: Coordinates the execution of all callbacks in the correct order and manages their interactions.
- `_TrainingState`: A dataclass summarizing all training parameters and serving as shared state across callbacks.

**Next**
I know that unittests are currently missing and that some callbacks might not work as expected in their current form. However, I would like to integrate the new MAPs architecture (PR #730 ) first and do the final Callbacks PR after that


clinicadl/callbacks/training_state.py
```diff
@@ -0,0 +1,71 @@
...
+    batch: int = -1
+
+    def reset(self, split: Split):
+        """Reset the training state for a new training split."""
+
+        self.n_batch = len(split.train_loader)
+        self.split = split
+        self.stop = False
+        self.epoch = 0
+        self.batch = 0
```
thibaultdvx, line 71: Just to understand: why the reset values are not the same as the default values?

clinicadl/metrics/metrics.py
```diff
@@ -32,37 +34,37 @@ class Metrics(ABC):
 
     @staticmethod
     def check_metrics(
-        metrics: list[Union[MetricConfig, MonaiMetric, CustomMetric, LossConfig, Loss]],
-    ) -> list[MetricConfig]:
+        metrics: Optional[dict[str, MetricType]],
```
thibaultdvx, line 37:
```suggestion
        metrics: dict[str, MetricType],
```
According to line 43, ``metrics`` can't be None.

clinicadl/metrics/metrics.py
```diff
@@ -32,37 +34,37 @@ class Metrics(ABC):
...
+            raise TypeError(
+                f"Metrics must be a dictionary, got {type(metrics)} instead."
+            )
 
-        if not isinstance(metrics, list):
-            metrics = [metrics]
-        for metric in metrics:
+        for metric_name, metric in metrics.items():
             if isinstance(metric, MonaiMetric):
                 config = get_metric_config(name=metric.__class__.__name__)
```
thibaultdvx: Doing this, you always take the default parameters of the metric, no matter what the user passed, don't you?

thibaultdvx (approved):
Great work, LGTM!

Regarding ``LRScheduler``, I think it's relevant to put it in the callbacks. But we should extend this class so that it accepts all the schedulers in ``clinicadl.optim.lr_schedulers``.

Besides, what do you think of putting all these callbacks in the ``__init__.py`` of ``callbacks``, so that the user can simply access callbacks via ``from clinicadl.callbacks import ...``?

clinicadl/metrics/metrics.py
camillebrianceau: I try something but i'll test it in the next PR !

clinicadl/callbacks/training_state.py
camillebrianceau: no idea

PR #729, opened 2025-07-02
CapsDataset tabular data + multiple tensor conversion

thibaultdvx:
Following last week's sprint:

- add a ``columns`` argument to ``CapsDataset``, to put some tabular data in the output ``DataPoint``;
- tensors are now save in ``.../t1_linear/tensors/{conversion_name}``, which enables to keep tensors from different conversion. By default (if no ``conversion_name`` is provided, will be saved in ``.../t1_linear/tensors/default``);
- update tests and documentation accordingly.


camillebrianceau (approved):
LGTM !

PR #786, opened 2025-07-21
Postprocessing transforms

thibaultdvx:
In this PR, I add some metrics for postprocessing. These metrics are from [MONAI](https://docs.monai.io/en/stable/transforms.html#post-processing), so I had to create a ``MonaiTransformWrapper`` so that they work as the transforms from TorchIO.

I also add parameters ``include`` and ``exclude`` to the transforms from TorchIO, so that the user doesn't necessarily have to apply the transform the all images in the ``DataPoint``.

**What's next?**

Modify the metrics so that they can apply postprocessing transform before the computation of the metric.


camillebrianceau (approved):
LGTM

PR #790, opened 2025-07-23
Metrics with postprocessing

thibaultdvx:
This PR aims to create a new ``Metric`` object that takes as input batches of ``DataPoints``.

Therefore, MONAI metrics are now wrapped in a ``MonaiMetricWrapper`` that converts them to ``Metric``.
This ``MonaiMetricWrapper`` also takes as arguments ``pred_key``, which is the key to the predictions in the input``DataPoints``, ``label_key``, which is the key to the ground truths in the input ``DataPoints``, and ``postprocessing``, which are potential postprocessing transforms to apply to the predictions and/or the ground truths before computing the metric.

``MetricConfigs`` were changed accordingly.

Finally, I fixed some very small issues with ``transforms``.


camillebrianceau (approved):
LGTM !

PR #791, opened 2025-07-24
Postprocessing

thibaultdvx:
- Make ``Transforms`` and ``Postprocessing`` inherits from the same ``TransformsHandler`` base class.
- Add ``exclude`` argument in ``MonaiTransformWrapper``, so that it behaves like TorchIO transforms.


camillebrianceau (approved):
LGTM! Just a small remark: I'm not entirely sure about the name `TransformHandler`, since our other handlers (like `CallbacksHandler` and `MetricsHandler`) are meant to manage and organize related objects (like `Callback` and `Metric`).

From what I understand of this class, maybe something like BaseTransform? WDYT?

PR #793, opened 2025-07-24
Callbacks and Trainer unittests

### LINKED ISSUE #735: Callbacks Unit Tests

### LINKED ISSUE #740: Trainer Unit Tests

camillebrianceau:
This PR should be ready for review:
- Added tests for callbacks
- Updated the callbacks themselves
- Added some documentation

The resume option seems to work, though I haven’t tested it yet with real data.


thibaultdvx:
Hi @camillebrianceau, I just changed trainer's tests so that the output MAPS are created in a temporary folder.

clinicadl/utils/enum.py
```diff
@@ -2,6 +2,8 @@
 
 
 class CaseInsensitiveEnum(str, Enum):
+    """ "Case insensitive Enum object."""
```
thibaultdvx:
```suggestion
    """Case insensitive Enum object."""
```

clinicadl/utils/pytorch_ssim.py
```diff
@@ -1,5 +1,6 @@
 ## From https://github.com/jinh0park/pytorch-ssim-3D
 
+# TODO : remove this or put it in Metrics module ?
```
thibaultdvx: I think we can delete this file as well. We now use the implementation of MONAI (that relies also on torch).

thibaultdvx (approved):
LGTM

PR #798, opened 2025-08-27
Test for gpu

thibaultdvx:
This PR would enable developers to add flags to test that requires a gpu or multiple gpus (via ``@pytest.mark.gpu`` or ``@pytest.mark.multi_gpu`` ).

Dedicated gpu-runners would then run these specific tests.


camillebrianceau (approved):
LGTM !

PR #799, opened 2025-08-29
Ddp test

thibaultdvx:
This PR aims to adding useful functions to test DDP frameworks. These functions are gathered in ``tests/utils.py``. And example of test with DDP is in ``tests/unittests/test_gpu.py``.


camillebrianceau (approved):
LGTM !

Issue #800, opened 2025-09-01
Extract Slices from TSV


HuguesRoy:
Good afternoon,  

I have a particular use case regarding slice extraction from images.  

Let’s say I want to train a model only on specific slices of each subject/session. For example, I have segmentation masks, and I want for subject A slice 85, subject B slice 95, and so on,  always in the axial direction.  

A TSV file describing slice indices for each subject/session:

```text
participant_id    session_id    slice_idx
sub_EXAMPLE01     ses-M000      85
sub_EXAMPLE02     ses-M000      80
sub_EXAMPLE03     ses-M000      35
```

The idea is to create a new extractor (SliceFromTSV) similar to the current Slice extractor, but instead of applying uniform rules (all slices, borders, discard, etc.), it uses a TSV mapping to select subject/session-specific slices.

This would be useful in settings where we want to train a model only on clinically relevant slices (e.g. tumor region, lesion location), or test a model on relevant slices.

### Proposed Solution

I implemented a prototype SliceFromTSV extractor that:

- Reads a TSV with (participant_id, session_id, slice_idx).
- Supports SliceDirection (sagittal, coronal, axial).

Has two modes:

`one_row_per_slice_mode`
- `True`: each TSV row corresponds to exactly one slice (recommended).
- `False`: multiple slices per subject/session, with internal handling of indices.

In order to compute the good length of the dataset object, I suggest to give the Datapoint object to the methods `num_samples_per_image` of the different extractor instead of the tensors directly, so it is more robust.
Why ? If I have the following tsv file:

```text
participant_id    session_id    slice_idx
sub_EXAMPLE01     ses-M000      85
sub_EXAMPLE02     ses-M000      80
sub_EXAMPLE02     ses-M000      35
```
The number of samples per image for the subject `sub_EXAMPLE01`is 1 but for the subject `sub_EXAMPLE02` it is 2.

PR #801, opened 2025-09-01
Add SliceFromTSV extraction

HuguesRoy:
PR for issue #800 — TSV-driven slice extraction

## Summary
This PR introduces a new extraction transform `SliceFromTSV` that selects subject/session-specific slice indices from a TSV file, and updates `CapsDataset` length computation to support TSV-driven per-subject slice counts.

## Motivation
Many workflows require training and evaluation on clinically relevant slices (e.g., tumor/lesion levels), which vary per subject/session. A TSV mapping `(participant_id, session_id, slice_idx)` makes this explicit and reproducible.

## What’s included
-  New `SliceFromTSV` extraction:
  - Reads a TSV with columns: `participant_id`, `session_id`, `slice_idx` (0-based).
  - Supports all `SliceDirection` values (sagittal/coronal/axial).
  - Two modes:
    - `one_row_per_slice_mode=True` (default): each TSV row is exactly one sample.
    - `False`: multiple slices per (subject, session); extractor exposes N samples for that datapoint.
  - Returns `SliceSample` with `slice_position`, `slice_direction`, `squeeze` consistent with existing `Slice`.
- CapsDataset: adjust dataset length logic to account for TSV-driven multiplicity (i.e., per-datapoint sample count).


thibaultdvx:
Hi @HuguesRoy, thanks a lot for this great PR. I took the liberty to add some minor modifications.
Besides, there are two points on which I would like your opinion:

## Merge SliceFromTSV with Slice

I think ``SliceFromTSV`` could be in ``Slice``. Currently, ``Slices`` accepts a selection of slices via the argument ``slices``, and it seems to me that your suggestion is very close to this behaviour.

I would thus suggest to add a new argument ``tsv_path`` (that would be mutually exclusive with ``slices``):

```python
class Slice(Extraction):

    def __init__(
        self,
        *,
        slices: Optional[List[NonNegativeInt]] = None,
        tsv_path: Optional[Union[str, Path]] = None,
        discarded_slices: Optional[List[NonNegativeInt]] = None,
        borders: Optional[Union[PositiveInt, Tuple[PositiveInt, PositiveInt]]] = None,
        slice_direction: SliceDirection = SliceDirection.SAGITTAL,
        squeeze: bool = True,
    ) -> None:
```

And then, I think, you just need to modify ``_get_slice_selection``, with something like that (not tested):

```python
def _get_slice_selection(self, data_point: DataPoint) -> np.ndarray[bool]:
        n_slices = data_point.image.tensor.size(self.slice_direction + 1)
        selection = np.ones(n_slices).astype(bool)

        if self.tsv_path:
                slice_ids = self._slices_for(data_point.participant, data_point.session)
        elif self.slices:
                slice_ids = self.slices

        if self.tsv_path or self.slices:
                # do the same as before when a slice selection was passed
                ...
```

## one_row_per_slice_mode?

I'm not sure to understand ``one_row_per_slice_mode``. Let's say I have a ``CapsDataset`` with two images, and a TSV like this:

```tsv
participant_id	session_id	slice_idx
sub-000	        ses-M000	1
sub-000	        ses-M000	3
sub-001	        ses-M000	1
```

Then, the length of the ``CapsDataset`` will be ``3``, because the first image has two slices selected, and the second one only one. When I call ``dataset[1]``, it means "get slice ``3`` from the image of ``(sub-000, ses-M000)``", and ``CapsDataset`` will call ``extract_sample(data_point, sample_index=1)``, because slice ``3`` is the second sample for this image.

In the same way, ``dataset[2]`` will call ``extract_sample(data_point, sample_index=0)``.

So, I don't see any problem if there are multiple rows or only one row for the same ``(subject, session)``. But, I probably misunderstand the purpose of this parameter... Can you enlighten me?

HuguesRoy:
Hi @thibaultdvx, thanks for the feedback!

I’m fine either adding this into `Slice` or keeping `SliceFromTSV` as a separate class. I initially split it out to avoid overloading `Slice` with TSV-specific logic, but I can move it if you prefer a single class ?

- With `one_row_per_slice_mode=True`:
  - takes only the first slice for `sub-000 ses-M000`, so dataset length = **2** (`sub-000 ses-M000 1` + `sub-001 ses-M000 1`)  

- With `one_row_per_slice_mode=False`:
  - The extractor groups by `(participant, session)` and uses all slices listed.  
  - `num_samples_per_image(...)` returns the number of slices requested for the current datapoint.  
  - Dataset length = sum of per-subject slice counts (e.g. 2 for `sub-000 ses-M000` + 1 for `sub-001 ses-M000` ⇒ **3** total).

It was in case you want to evaluate on only one slice per subject without rewriting the TSV, but we could only use the behavior `one_row_per_slice_mode=False`

thibaultdvx:
Hi @HuguesRoy, thanks for the clarification.

I think it's completely ok to put it in ``Slice``, but I would prefer that we do without the parameter ``one_row_per_slice_mode``. I think that it would make the class more difficult to understand without offering a significant advantage for the user.

thibaultdvx:
Hi @HuguesRoy, I added 3 reviews, on which I would like to have your opinion:

- ``1st commit``: I guess it was not done on purpose, but you removed all the docstrings; so, I put them back. Besides, for more clarity, I removed some minor changes that were out of the scope of this PR. Finally, I made other minor changes to be consistent with ClinicaDL.
- ``2nd commit``:  you made the argument ``tsv_path`` to have "full precedence", but I think it's better to raise an error if mutually exclusive arguments are called, instead of ignoring some. So, I made it impossible for the user to use ``tsv_path`` with ``slices`` for example.
- ``3rd commit``: I saw that you had some troubles to modify ``Slice`` because of the way the base class ``Extraction`` was implemented. So, I modified ``Extraction`` so that the modifications needed in ``Slice`` are much lighter. Tests are passing, so I guess this new version satisfies what you wanted to do.

PR #802, opened 2025-09-02
New batch object

thibaultdvx:
Some updates on the ``SimpleBatch`` object:

- renamed to ``Batch``;
- ``get_images`` and ``get_labels`` have been merged to a common method ``get_field``, that enables to access any field of the ``DataPoint`` in the batch;
- a new ``to`` method that enables to send the batch on an other device.

Tests and documentation have been updated accordingly.


camillebrianceau (approved):
LGTM !

PR #811, opened 2025-09-22
Lr scheduler

### LINKED ISSUE #805: LRScheduler ``.step()`` location inaccurate
https://github.com/aramis-lab/clinicadl/blob/83d753494afd0fd33adc8683c6012eefcf482983/clinicadl/callbacks/factory/lr_scheduler.py#L163

Some LRSchedulers expects the ``scheduler.step()`` after every optimization steps (e.g. [CyclicLR](https://docs.pytorch.org/docs/stable/generated/torch.optim.lr_scheduler.CyclicLR.html#cycliclr)), whereas others expect it at the end of the epoch (e.g. [StepLR](https://docs.pytorch.org/docs/stable/generated/torch.optim.lr_scheduler.StepLR.html#steplr)).

thibaultdvx:
Fixes #805

Also added methods ``save_checkpoint`` and ``load_checkpoint`` to the callback ``LRScheduler``.


camillebrianceau (approved):
LGTM ! Just a quick point: shouldn’t you check the optimizer consistency at the beginning of training to ensure that the optimizer passed to the model matches the one used by the scheduler?

thibaultdvx:
> LGTM ! Just a quick point: shouldn’t you check the optimizer consistency at the beginning of training to ensure that the optimizer passed to the model matches the one used by the scheduler?

Well we can try to implement these kind of checks, but the thing is that ``ClinicaDLModel`` doesn't necessarily have an attribute ``optimizer``, e.g. if there are several optimizers.

So, I would say we can check if the model has an attribute named optimizer, and raise a warning if it differs from the one passed in the lr scheduler.

Another option, is to add a method named ``get_optimizers`` to ``ClinicaDLModel``, and use these methods to retreive the optimizers when we need them. Besides, this method could also be helpful for Tensorboard, to save all the learning rates during training.

thibaultdvx:
I just implemented the ``get_optimizers`` option, with a check in ``on_train_begin`` of ``LRScheduler``.

PR #821, opened 2025-11-07
Fix pattern in custom preprocessing file type

HuguesRoy:
Changed folder name to `custom` using the `PreprocessingMethod` enum.


thibaultdvx:
Hey @HuguesRoy, thanks for the contribution!

Could you please modify the test accordingly?

https://github.com/aramis-lab/clinicadl/blob/a749df4499169494649372e2228646ff091b7e1a/tests/unittests/data/datatypes/test_preprocessing.py#L18

